PR #32, opened 2023-06-07
Update requirements to enable circuit prints

David-Kreplin:


MoritzWillmann:
I'm not sure if we should add matplotlib to the dependencies since we don't use it anywhere in the source code.

What about a new optional dependency entry like
```yaml
examples = ["jupyter", "matplotlib>=3.5", "pylatexenc>=2.10"]
```
that holds all necessary dependencies to run the examples?

Then one could run
```bash
pip install squlearn[examples]
```
to install it with said dependencies.

Not sure if the name `examples` might be confusing though...

David-Kreplin:
We have this feature_map.draw() command that explicitly uses the matplotlib option of qiskit. 
But qiskit will also raise an error with the option to install matplotlib, that is why it might not be necessary. 
The example option sounds interessting, let's discuss it in the next meeting.

PR #33, opened 2023-06-07
format code & add action

MoritzWillmann:
This PR formats all existing code and adds a github action to check formatting in PRs


David-Kreplin (commented):
The new standard changes 'a' -> "a" is that really good?
Probably a matter of taste...

How does the format.yml file work?
I see "runs-on: ubuntu-latest" is this run on github or locally?

MoritzWillmann:
It's the black standard so I'd suggest we use it.
The action runs on github in some container when someone opens a PR.

David-Kreplin:
The container is fantastic!
Do you now, how docstrings are affected by that?
That's where I sometimes would need the ' ticks.

David-Kreplin (approved):

PR #34, opened 2023-06-12
Fix of the QNNC program for 1D outputs

David-Kreplin:


MoritzWillmann (approved):
looks good, also black check seems to work 🎉

PR #39, opened 2023-06-15
add some good names for pylint

MoritzWillmann:


David-Kreplin (approved):
Good idea! I don't see a problem using single character variables.
In scientific computing, this makes things more simple.

PR #45, opened 2023-06-16
Flit build system

MoritzWillmann:
- Change buildsystem to [flit](https://flit.pypa.io/en/stable/index.html).
- Add workflow to publish to PyPI on release


MoritzWillmann:
I just tested it in a separate repo with identical settings. Should work 😄

PR #49, opened 2023-06-28
Refactor of the loss class

David-Kreplin:


src/squlearn/qnn/loss.py
```diff
@@ -2,64 +2,316 @@
 import abc
 from typing import Union
 import numpy as np
+from collections.abc import Callable
 
 
 class LossBase(abc.ABC):
-    """Base class implementation for Loss functions.
+    """Base class implementation for Loss functions."""
```
MoritzWillmann, line 9:
```suggestion
    """Base class implementation for loss functions."""
```

src/squlearn/qnn/loss.py
```diff
@@ -2,64 +2,316 @@
...
+        Args:
+            opt_param_op (bool): True, if operator has trainable parameters
+        """
+        self._opt_param_op = opt_param_op
 
     @property
     @abc.abstractmethod
     def loss_args_tuple(self) -> tuple:
-        """Returns evaluation tuple for loss calculation.
+        """Returns evaluation tuple for Loss calculation."""
```
MoritzWillmann, line 25:
```suggestion
        """Returns evaluation tuple for loss calculation."""
```

src/squlearn/qnn/loss.py
```diff
@@ -2,64 +2,316 @@
...
+        raise NotImplementedError()
 
-        Returns:
-            Evaluation tuple
-        """
+    @property
+    @abc.abstractmethod
+    def gradient_args_tuple(self) -> tuple:
+        """Returns evaluation tuple for Loss gradient calculation."""
```
MoritzWillmann, line 31:
```suggestion
        """Returns evaluation tuple for loss gradient calculation."""
```

src/squlearn/qnn/loss.py
```diff
@@ -2,64 +2,316 @@
...
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def value(self, value_dict: dict, **kwargs) -> float:
+        """Calculates and returns the Loss value."""
```
MoritzWillmann, line 36:
```suggestion
        """Calculates and returns the loss value."""
```

src/squlearn/qnn/loss.py
```diff
@@ -2,64 +2,316 @@
...
         raise NotImplementedError()
 
-    @staticmethod
     @abc.abstractmethod
-    def gradient_args_tuple(opt_param_op: bool = True) -> tuple:
-        """Returns evaluation tuple for gradient calculation.
+    def gradient(
+        self, value_dict: dict, **kwargs
+    ) -> Union[np.ndarray, tuple[np.ndarray, np.ndarray]]:
+        """Calculates and returns the gradient of the Loss."""
```
MoritzWillmann, line 43:
```suggestion
        """Calculates and returns the gradient of the loss."""
```

src/squlearn/qnn/loss.py
```diff
@@ -2,64 +2,316 @@
...
 
-    @staticmethod
-    @abc.abstractmethod
-    def gradient(value_dict: dict, **kwargs) -> Union[np.ndarray, tuple[np.ndarray, np.ndarray]]:
-        """Calculates and returns the gradient value."""
-        raise NotImplementedError()
+    def gradient(
+        self, value_dict: dict, **kwargs
+    ) -> Union[np.ndarray, tuple[np.ndarray, np.ndarray]]:
+        """Calculates and returns the gradient of the composed Loss.
```
MoritzWillmann, line 187:
```suggestion
        """Calculates and returns the gradient of the composed loss.
```

src/squlearn/qnn/loss.py
```diff
@@ -133,77 +386,82 @@ def gradient(value_dict: dict, **kwargs) -> Union[np.ndarray, tuple[np.ndarray,
...
+    def loss_args_tuple(self) -> tuple:
+        """Returns evaluation tuple for loss calculation."""
+        return ("var",)
 
-    @staticmethod
-    def gradient_args_tuple(opt_param_op: bool = True) -> tuple:
-        if opt_param_op:
+    @property
+    def gradient_args_tuple(self) -> tuple:
+        """Returns evaluation tuple for Loss gradient calculation."""
```
MoritzWillmann, line 400:
```suggestion
        """Returns evaluation tuple for loss gradient calculation."""
```

MoritzWillmann (commented):
Thoughts:
- Genreally it looks really good. It totally makes sense with all the math operations.
- Do we want to integrate the `VarianceLoss` inside of the `QNNRegressor` and `QNNClassifier`? In terms of combinability and consistency (with not yet existent other losses), I'd prefer if this was left to the user to do. Also the interface is already really cramped and adding more and more params just makes things worse.
- I know it's also in my code (but why not fix it now?), we should spell `loss` not `Loss` in the docstrings.

David-Kreplin:
Thanks for fixing the docstrings!

Regarding the VarianceLoss: My changes already simplify a lot! In the current version, you have to parse a list of losses to realize the variance regularization. That was not an intuitive solution in my opinion. 
The variance regularization as a hyperparameter was implemented before, and I think, it also should stay, since this is a feature we want to highlight. 
The fact that the loss is techically modified doesn't have to bother the user.

MoritzWillmann:
You're right, the `variance` argument was there already and it's a bit more intuitive now. Still don't you think using the loss like
```python
QNNRegressor(...,SquaredLoss() + VarianceLoss(some_variance_factor),...)
```
or even
```python
loss = SquaredLoss() + VarianceLoss(some_variance_factor)
QNNRegressor(...,loss,...)
```
could be more intuitive than
```python
QNNRegressor(...,SquaredLoss(),...,variance=some_variance_factor,...)
```
Especially when we add other Losses we might combine it with. Then the whole loss construction could take place at once.

src/squlearn/qnn/loss.py
```diff
@@ -131,79 +384,88 @@ def gradient(value_dict: dict, **kwargs) -> Union[np.ndarray, tuple[np.ndarray,
 
 
 class VarianceLoss(LossBase):
-    """Variance loss for regression."""
+    """Variance loss for regression.
```
MoritzWillmann, line 387:
```suggestion
    r"""Variance loss for regression.
```

David-Kreplin:
I think your first two approaches are more complicated, since the user has to know exactly what he/she is doing.
In the last interface, it is a hyper parameter of the method which might be more intuitive. 
Nevertheless, the first implementation should already work, if a user wants to do it that way.

David-Kreplin:
The "feature" might also be overlooked if it is not a hyper-parameter.

MoritzWillmann (approved):
Alright

PR #54, opened 2023-07-04
fix for no parameters in the projected quantum kernel program

David-Kreplin:


JanSchnabel (approved):
Nice, looks good, thanks for fixing this issue!

PR #60, opened 2023-07-13
Add documentation and tutorial for quantum feature maps

David-Kreplin:


David-Kreplin:
Basically all modifications are doc strings. There are also some improvements in the feature maps options.

MoritzWillmann:
How close do we want to stick to sklearn docstrings? If close, then we might need to change the order in some of the sections (i.e. bringing args up to after the long description and moving examples to the bottom) aswell as adding a separate reference section instead of having it in the first line description.

David-Kreplin:
Don't know if that is really necessary. 
At the moment, I'm happy to reach a level that is usefull but not a massive amount of work.
If the package goes high, we can rethink the documentation in a more structured way.

rothma (approved):

PR #65, opened 2023-07-19
Add low level docstrings to kernel base classes

JanSchnabel:


David-Kreplin:
Black says the file "qkrr.py" is not correctly formated.

David-Kreplin (approved):

PR #66, opened 2023-07-19
Add num_layer parameter to LayeredFeatureMap.from_string

rothma:


David-Kreplin:
You know that you can achieve this by the squared brackets? "N[...]"

rothma:
> 

I do. I think it still makes sense:

1.  If I use the FM created in the example notebook and I try to create multiple layers of the FM via N[] it does not work, because there is already a single block repetition within the FM and the function does not seem to be able to process multiple M[]-style repetition.
2. I think it's intuitive from a user perspective. I tried to reduce the redundancy be framing the N[]-method as to be used for single blocks within the FM and the num_layers as a way to easily repeat everything.

1. is fixable and a borderline issue by itself, but I think 2. is still valid even if 1 is fixed.

David-Kreplin:
Ok, I don't mind. 
But the black check is failing, which can be resolved by running black in the terminal for the file layered_feature_map.py.

rothma:
> Black test is still failing for layered_feature_map.py :/

I've seen it. Too bad the error message does not give any information about why it fails. I ran black on the file...

MoritzWillmann:
Sorted 😄

David-Kreplin (approved):

PR #70, opened 2023-07-19
Add qsvm documentation

rothma:
coef_ property currently still in generated sphinx documentation. Should be removed in future updates


src/squlearn/kernel/ml/qsvr.py
```diff
@@ -9,13 +9,52 @@ class QSVR(SVR):
 
     This class is a wrapper of sklearn.svm.SVR. It uses a quantum kernel matrix
```
MoritzWillmann, line 10: Could you change this paragraph, so it wraps to less than 99 characters per line? Black doesn't do this for docstrings...

MoritzWillmann (changes requested):

MoritzWillmann (approved):
Looks good!

PR #82, opened 2023-07-26
Refactor regularization to kernel matrix base class

rothma:
- Removed regularization from high-level classes and moved to KernelMatrixBase (qgpr is a bit of an exception)
- Added regularization to QSVR and QSVC

**JSL, DKR and FAP need to approve before merging. JSL & FAP please verify that your high-level methods still look fine.**


src/squlearn/kernel/matrix/projected_quantum_kernel.py
```diff
@@ -429,7 +437,10 @@ def evaluate(self, x: np.ndarray, y: np.ndarray = None) -> np.ndarray:
         if self._parameters is None:
             raise ValueError("Parameters have not been set yet!")
 
-        return self._outer_kernel(self._qnn, self._parameters, x, y)
+        kernel_matrix = self._outer_kernel(self._qnn, self._parameters, x, y)
+        if regularization is not None:
+            kernel_matrix = self._regularize_matrix(kernel_matrix, method=regularization)
```
rothma, line 442: I assume this is the correct spot to put the regularization?

src/squlearn/kernel/ml/qsvc.py
```diff
@@ -60,9 +66,16 @@ class QSVC(SVC):
...
+            self,
+            *,
+            quantum_kernel: KernelMatrixBase,
+            regularization: Union[str, None] = None,
+            **kwargs) -> None:
         self.quantum_kernel = quantum_kernel
-        super().__init__(kernel=self.quantum_kernel.evaluate, **kwargs)
+        super().__init__(
+            kernel=lambda x, y: self.quantum_kernel.evaluate(x, y,
+            regularization=regularization), **kwargs)
```
rothma, line 78: too hacky?

JanSchnabel (commented):
Do we need the "regularization" argument also in evaluate() of the _ComposedKernelMatrix class?

rothma:
> Do we need the "regularization" argument also in evaluate() of the _ComposedKernelMatrix class?

I'd like to raise the question if we should get rid of the _ComposedKernelMatrix class altogether. Or does someone ever use it?

David-Kreplin:
I'm not a big fan of 

> 


I strongly disagree :D 
I think no one utilizes that in the moment (probably no one is aware that this exists), but when we really want to solve difficult (Q)ML problems, we should implement features like that.

First, it is something that other QML tools very likely don't provide, but is implemented in all sklearn classes. 

Second, it is usually straight forward to implement and maintain (here: just add regularize in the subcalls)

Third, it allows the combination of multiple feature maps or even combinations of projected and fidelity kernels. As far as I'm aware, that is the stuff people usually do, when they handcraft more complicated ML methods.

rothma:
> > I'd like to raise the question if we should get rid of the _ComposedKernelMatrix class altogether. Or does someone ever use it?
> 
> I strongly disagree :D I think no one utilizes that in the moment (probably no one is aware that this exists), but when we really want to solve difficult (Q)ML problems, we should implement features like that.
> 
> First, it is something that other QML tools very likely don't provide, but is implemented in all sklearn classes.
> 
> Second, it is usually straight forward to implement and maintain (here: just add regularize in the subcalls)
> 
> Third, it allows the combination of multiple feature maps or even combinations of projected and fidelity kernels. As far as I'm aware, that is the stuff people usually do, when they handcraft more complicated ML methods.

Okay, I get at least some of your points. My (provocative) question about deletion came from the more fundamental question about wether we should have features that are not actively used but are exist „just in case“? The downside is that you have to maintain a bunch of stuff that which might never get actively used but still requires work.

src/squlearn/kernel/ml/qsvc.py
MoritzWillmann: I don't think so. PEP8 favors `def`` over `lambda` but I like it 😄

David-Kreplin:
What is your opinon on comment on putting the ``regularization=True`` into the ``__init__`` instaead of the ``evaluate`` function?

David-Kreplin:
Yes exactely, for me that sounds more intuitive, but I would like to here the opinion of @JanSchnabel and @FredericRapp on that.

JanSchnabel:
> Yes exactely, for me that sounds more intuitive, but I would like to here the opinion of @JanSchnabel and @FredericRapp on that.

To me both make sense. The regularization techniques we are talking about refer strictly speaking to the kernel and not to the respective high-level algorithms (they have their own regularization). Moreover, putting the `regularization=True` into the `__init__`, would harmonize the `mit_depol_noise` argument which is also in the `__init__` of the `FidelityQuantumKernel` class

rothma:
> Yes exactely, for me that sounds more intuitive, but I would like to here the opinion of @JanSchnabel and @FredericRapp on that.

okay, done.

JanSchnabel (commented):
I agree with the changes and particularly like the "I am threshoding" and "I am tikhonovin" print statements :D 
However, I have one question: Do we need to add the `regularization` parameter to `get_params()` and `set_params()` in the respective kernel methods? ?

rothma:
> I agree with the changes and particularly like the "I am threshoding" and "I am tikhonovin" print statements :D However, I have one question: Do we need to add the `regularization` parameter to `get_params()` and `set_params()` in the respective kernel methods?

Good catch with the prints. I used them for testing. I took them out now.

As to wether they need to be included in the setters and getters comes to the choice of whether we want to be able to change it in the high-level methods. I don't have a strong oppinion about that but I have a slight tendency for no.

JanSchnabel (approved):

PR #83, opened 2023-07-26
Add default parameter initialization (affects all low-level methods)

David-Kreplin:


rothma:
Having specific bounds for every feature map adds quite some complexity for new developments and maintenance. Do you see any way of factorin this into some lower level?

src/squlearn/expectation_operator/expectation_operator_base.py
```diff
@@ -53,6 +53,33 @@ def num_parameters(self):
...
+
+        Args:
+            ones (bool): If True, returns an array of ones (default: True)
+            seed (Union[int,None]): Seed for the random number generator
+
+        Return:
+            The randomly generated parameters
+        """
+        if ones:
+            return np.ones(self.num_parameters)
```
rothma, line 75: else:

src/squlearn/expectation_operator/expectation_operator_base.py
```diff
@@ -53,6 +53,33 @@ def num_parameters(self):
...
+
+        if self.num_parameters == 0:
+            return np.array([])
```
rothma, line 78: else:

src/squlearn/expectation_operator/expectation_operator_base.py
David-Kreplin: Interestingly, pylint suggests usually the opposit, hence:

```python
if a=b:
     return a

# continue with code 
``` 

instead of 

```python
if a=b:
     return a
else:
     # continue with code 
```

David-Kreplin:
> Having specific bounds for every feature map adds quite some complexity for new developments and maintenance. Do you see any way of factorin this into some lower level?

Not really, since we can have feature maps, where the parameter intilization depends on the structure of the feature map.
E.g. the parameters of the chebychev encoding are initilized with fixed values and not randomly (starting from 0 going to a predefined hyper-parameter along the qubits)
Also I have seen that parameters of the two qubit rotation gates can show a larger periodicity of 4pi instead of 2pi.
So the inital parameters depend on the feature map structure. 

We can add a default boundary of (-pi,pi) so if nothing is implemented, these are the values.

src/squlearn/expectation_operator/expectation_operator_base.py
rothma: Okay. To me, the latter is more readable but if there‘s a convention it makes sense to follow it.

rothma:
> > Having specific bounds for every feature map adds quite some complexity for new developments and maintenance. Do you see any way of factorin this into some lower level?
> 
> Not really, since we can have feature maps, where the parameter intilization depends on the structure of the feature map. E.g. the parameters of the chebychev encoding are initilized with fixed values and not randomly (starting from 0 going to a predefined hyper-parameter along the qubits) Also I have seen that parameters of the two qubit rotation gates can show a larger periodicity of 4pi instead of 2pi. So the inital parameters depend on the feature map structure.
> 
> We can add a default boundary of (-pi,pi) so if nothing is implemented, these are the values.

What about defining classes, e.g. 1-qubit and 2-qubit gates which the fm can auto

> > Having specific bounds for every feature map adds quite some complexity for new developments and maintenance. Do you see any way of factorin this into some lower level?
> 
> Not really, since we can have feature maps, where the parameter intilization depends on the structure of the feature map. E.g. the parameters of the chebychev encoding are initilized with fixed values and not randomly (starting from 0 going to a predefined hyper-parameter along the qubits) Also I have seen that parameters of the two qubit rotation gates can show a larger periodicity of 4pi instead of 2pi. So the inital parameters depend on the feature map structure.
> 
> We can add a default boundary of (-pi,pi) so if nothing is implemented, these are the values.

Do I understand correctly that the challenge is to set these bounds without looking at (generating) the circuit? Because if we could look at it, we could define classes, e.g. singlequbit and twoqubit, nomlinear) gates where the first two all have the same bounds. The fm could then iterate through the circuit and only treat the exceptions (like non-linear encodings) explicitly. Then one would only need to modify the bound property for fms with actual exceptions and could treat the rest in the base class.

David-Kreplin:
> > > Having specific bounds for every feature map adds quite some complexity for new developments and maintenance. Do you see any way of factorin this into some lower level?
> > 
> > 
> > Not really, since we can have feature maps, where the parameter intilization depends on the structure of the feature map. E.g. the parameters of the chebychev encoding are initilized with fixed values and not randomly (starting from 0 going to a predefined hyper-parameter along the qubits) Also I have seen that parameters of the two qubit rotation gates can show a larger periodicity of 4pi instead of 2pi. So the inital parameters depend on the feature map structure.
> > We can add a default boundary of (-pi,pi) so if nothing is implemented, these are the values.
> 
> Do I understand correctly that the challenge is to set these bounds without looking at (generating) the circuit? Because if we could look at it, we could define classes, e.g. singlequbit and twoqubit, nomlinear) gates where the first two all have the same bounds. The fm could then iterate through the circuit and only treat the exceptions (like non-linear encodings) explicitly. Then one would only need to modify the bound property for fms with actual exceptions and could treat the rest in the base class.

Yes that is something I will let Andre implement in a simple form for the Layered Approach.
The problem in general is, that the bounds depend on the construction of the feature map, especially for non-linear encodings we can't tell the boundaries without knowing the design. 
The only thing that has to be implemented for a feature map is an extra function that gives the bounds. Personally, I don't find this a large overhead. 
But I will add default boundaries of (-pi,pi) in the base class, so it will work if one forgets about it.

rothma (approved):

PR #90, opened 2023-08-04
Add get_params() and set_params() to QKRR

JanSchnabel:
* Can't set "gamma" parameter when using "gaussian" outer_kernel in PQK => didn't manage to fix this
* Modified FQK such that evaluate_duplicates and mit_depol_noise is also returned from get_params(deep=True) -> not sure if we really want this?
* Do we want to include the regularization parameter into getters and setters?

### DIFF
--- a/src/squlearn/kernel/matrix/fidelity_kernel.py
+++ b/src/squlearn/kernel/matrix/fidelity_kernel.py
@@ -134,7 +134,7 @@ class FidelityKernel(KernelMatrixBase):
         params["evaluate_duplicates"] = self._evaluate_duplicates
         params["mit_depol_noise"] = self._mit_depol_noise
         if deep:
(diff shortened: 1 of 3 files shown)


David-Kreplin:
> * Modified FQK such that evaluate_duplicates and mit_depol_noise is also returned from get_params(deep=True) -> not sure if we really want this?

Thanks for fixing that, forgot about that issue

David-Kreplin:
> * Can't set "gamma" parameter when using "gaussian" outer_kernel in PQK => didn't manage to fix this

very strange, I will take a look.

JanSchnabel:
> 

Yes, exactly. I can take care of this

David-Kreplin (approved):
kwargs, set_params and get_params now working

PR #92, opened 2023-08-08
A couple of smaller convenience fixes 

David-Kreplin:
- Executor default set to statevector simulator
- Fix in set_params for PQK (no recalculation of the kernel matrix necessary when changing outer parameters)
- Corrections for QKRR set_params
- Decompose option for QiskitFeatureMap


src/squlearn/feature_map/feature_map_implemented/qiskit_feature_map.py
JanSchnabel, line 1: Do we want to add `decompose` and `kwargs` to the docstrings?

src/squlearn/kernel/ml/qkrr.py
```diff
@@ -160,15 +158,12 @@ def set_params(self, **params):
                 )
 
         param_dict = {}
+        valid_params = self.get_params().keys()
```
rothma, line 161: this is the same dict for the same check again

David-Kreplin: Yes that was a bug, that is fixed in the last commit

src/squlearn/kernel/ml/qkrr.py
```diff
@@ -160,15 +158,12 @@ def set_params(self, **params):
...
         for key, value in params.items():
-            # if key in valid_params:
-            if key in self._quantum_kernel.get_params().keys():
+            if key in valid_params:
                 param_dict[key] = value
-        self._quantum_kernel.set_params(**param_dict)
+        if len(param_dict) > 0:
+            self._quantum_kernel.set_params(**param_dict)
```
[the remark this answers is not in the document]
David-Kreplin: Yes that was a bug, that is fixed in the last commit

Issue #62, opened 2023-07-18
High-level methods need to be able to set feature map hyperparameters (e.g. num_qubits, num_layers)


rothma:
- Problem could e.g. arise in ProjectedKernel which already create a circuit when initialized (which can happen before it is fed into a high-level ML method)

rothma:
Same for projected kernel length-scale parameter?

MoritzWillmann:
Implemented in #90 #99 and #107

PR #99, opened 2023-08-15
Add `get_params` and `set_params`  for qnns

MoritzWillmann:
This PR

- adds a reset function for `SGDMixin` optimizers
- does some refactoring of `BaseQNN`'s parameters
- adds `get_params` and `set_params` for `BaseQNN`
- adds tests for `get_params` and `set_params` in `BaseQNN`

partially solves #62


src/squlearn/qnn/base_qnn.py
```diff
@@ -108,11 +116,132 @@ def fit(self, X: np.ndarray, y: np.ndarray, weights: np.ndarray = None) -> None:
...
+                    for key, value in oper_dict.items():
+                        if key != "num_qubits":
+                            params[f"op{i}__{key}"] = value
+            else:
+                for key, value in self._operator.get_params().items():
+                    if key != "num_qubits":
+                        params[f"operator__{key}"] = value
+        return params
+
+    def set_params(self: BaseQNN, **params) -> BaseQNN:
```
David-Kreplin, line 151: Short question: Why are the feature map and the operator are stored and treated separately from the QNN? Aren’t this parameter showing up twice? get_params and set_params of the QNN automatically update the featuremap and the operator stored within the QNN and reinitialize the QNN if necessary. Not sure, why it is necessary to change (or even store) the operator and the feature map in the baseclass.

MoritzWillmann: You're right. This way it's way easier. I changed it accordingly.

src/squlearn/qnn/base_qnn.py
```diff
@@ -15,12 +18,14 @@
 from .qnn import QNN
 from .training import shot_adjusting_options
 
+MULTI_OP_KEY_PATTERN = re.compile(r"op(\d+)__(.*)")
```
David-Kreplin, line 21: for what is that needed?

src/squlearn/qnn/base_qnn.py
```diff
@@ -108,11 +121,85 @@ def fit(self, X: np.ndarray, y: np.ndarray, weights: np.ndarray = None) -> None:
...
+
+        # Set parameters of the QNN
+        qnn_params = params.keys() & self._qnn.get_params(deep=True).keys()
+        if qnn_params:
+            self._qnn.set_params(
+                **{key: value for key, value in params.items() if key in qnn_params}
+            )
+
+            # If the number of parameters has changed, reinitialize the parameters
+            if self.feature_map.num_parameters != len(self.param_ini):
```
David-Kreplin, line 179: not sure if that is working, since only the qnn is changed, but maybe the self.feature_map and self._qnn.self_feature_map are the same object

src/squlearn/qnn/base_qnn.py
```diff
@@ -108,11 +121,85 @@ def fit(self, X: np.ndarray, y: np.ndarray, weights: np.ndarray = None) -> None:
...
+            if isinstance(self.operator, list):
+                num_op_parameters = sum(operator.num_parameters for operator in self.operator)
+                if num_op_parameters != len(self.param_op_ini):
+                    self.param_op_ini = np.concatenate(
+                        [
+                            operator.generate_initial_parameters(seed=self.parameter_seed)
+                            for operator in self.operator
+                        ]
+                    )
+            elif self.operator.num_parameters != len(self.param_op_ini):
```
David-Kreplin, line 192: same here, you can get the number of parameters and operator_parameters also from the qnn class.

src/squlearn/qnn/base_qnn.py
MoritzWillmann: Leftover, thanks for spotting

src/squlearn/qnn/base_qnn.py
MoritzWillmann: They are the same

David-Kreplin (approved):

PR #119, opened 2023-08-24
fix sklearn parameter searches

rothma:


rothma:
- add return self to all set_params methods of all kernel algorithms
- do not try to regularize non-square (train-test) gram matrices

MoritzWillmann (approved):
🎉

PR #139, opened 2023-09-07
Add user guide qnn

MoritzWillmann:
This is my initial suggestion, I'm happy to take any kind of notes.

This PR solves #115


rothma:
What's the status of this PR?

MoritzWillmann:
> What's the status of this PR?

It would be good if someone could proofread this. I only set it to WIP so it doesn't get merged yet

David-Kreplin:
I will take care of this next week, since I was originally assigned to this task.

David-Kreplin:
I added some first edits untio the optimization section

David-Kreplin:
I reviewed and edited the qnn user guide:

 - Add an example how to initilize the QNN class
 - Changed the plot from svg to dynamically constructed once
 - Overall edits and corrections of what has been written. 
 
From my side, that is finished and can be merged.

David-Kreplin:
Unfortunatly referencing the figures does not work for the matplotlib output. I did some research, but couldn't figure out how that is done, so I just changed the caption and the reference in the text

MoritzWillmann:
> Unfortunatly referencing the figures does not work for the matplotlib output. I did some research, but couldn't figure out how that is done, so I just changed the caption and the reference in the text

This still looks a bit weird, I'll try to fix it. Honestly I'm not entirely sure if I like the plot directive too much...

David-Kreplin:
> > Unfortunatly referencing the figures does not work for the matplotlib output. I did some research, but couldn't figure out how that is done, so I just changed the caption and the reference in the text
> 
> This still looks a bit weird, I'll try to fix it. Honestly I'm not entirely sure if I like the plot directive too much...

Great thing about the plot directive is, that is done dynamically, hence when things are changing, wo don't have to produce another svg manually. Better improvement would be, if all plot / python codes are in files, so that search and replace / refactor would work if we rename stuff.

MoritzWillmann:
I just fixed the numbering of the captions. From my side everything looks good 🎉

PR #146, opened 2023-09-20
Improve kwargs explanation in docu

rothma:
Related to issue https://github.com/sQUlearn/squlearn/issues/136


JanSchnabel (commented):
I like the changes. For convenience, we could consider to keep the links to the respective sklearn documentation in QSVC and QSVR?

rothma:
> I like the changes. For convenience, we could consider to keep the links to the respective sklearn documentation in QSVC and QSVR?

apperently moritz did some sphinx magic and now the sklearn (and qiskit) documentation is linked. This means that the sklearn.svm.SVR is clickable and leads to the respective sklearn documentation website. Therefore, I have removed the explicit links.

JanSchnabel (approved):

Issue #11, opened 2023-05-28
Reiterate Example Notebooks


MoritzWillmann:
This includes:
- checking for necessity
- checking if runs
- adding more explanations

For
- [x] `executor`
- [x] `feature_maps`
- [x] `kernel`
- [x] `qnn`
- [x] `tutorials`

rothma:
This is done, isn't it?

JanSchnabel:
> This is done, isn't it?

For Executor and feature maps, there is a tick missing

David-Kreplin:
I finished the remaining examples and close now this issue.

PR #149, opened 2023-09-26
Reiterate QNN Examples

MoritzWillmann:
partially solves #11


examples/qnn/expectation_operator.ipynb
David-Kreplin, line 1: Why removing this example?
It is the only one displaying the functionality of the expection operator class

examples/qnn/example_minibatch.ipynb
David-Kreplin, line 1: You strongly reduced the epochs, I assume to improve runtime for this example which makes sense. 
Wouldn't it be better to start from preoptimized parameters, so that the final result looks nicer?

examples/qnn/regression_example.ipynb
David-Kreplin, line 1: Any chance that we can remove the: C:\Users\mow\Documents\AutoQML\squlearn\src\squlearn\qnn\qnnr.py:133: UserWarning: The model is not fitted.
  warn("The model is not fitted.") warning

MoritzWillmann: Done 🙂 A bit of a hack but it works

examples/qnn/example_minibatch.ipynb
MoritzWillmann: Done 🙂

David-Kreplin (approved):

Issue #112, opened 2023-08-23
QNN Class has no classification loss


rothma:
This also needs to be changed in the QNN classification example

PR #154, opened 2023-09-27
Add `CrossEntropyLoss`

MoritzWillmann:
## Current State

This PR adds a new loss class `CrossEntropyLoss` implementing the multi-class log loss (aka. cross entropy loss)

The value is calculated as

$$L = -\frac{1}{N} \sum_{i=1}^N \sum_{j=1}^K y_{ij} \cdot \log\left(f(x_i)_j\right)$$

If the task at hand is binary, a "class" is added with prediction $1-f(x)$ and label $1-y$.

The gradient is thus calculated as

$$\frac{\partial}{\partial \theta} L = -\frac{1}{N}\sum_{i=1}^N \sum_{j=1}^K \frac{y_{ij}}{f(x_i)_j} \frac{\partial f}{\partial \theta}$$

If the task at hand is binary, a "class" is added with prediction $f(x)-1$ and label $1-y$.

I test this functionality in `tests/qnn/test_loss.py`.

This PR solves #112


MoritzWillmann:
## Update
I figured out that if if `eps` is sufficiently large, training works. The idea of `eps` being dependent on the learning task and model at hand, i.e. making it somewhat of a hyperparameter, seems very strange to me. Maybe we can gain some more insights on this.

MoritzWillmann:
Not sure, why this was closed

rothma:
@MoritzWillmann in my tests it works now (at least for binary)

Issue #144, opened 2023-09-19
Allow for precomputed quantum kernels in kernel methods


JanSchnabel:
@David-Kreplin and @MoritzWillmann we discussed this at 19.09.23. If you agree on the suggested feature, I will implement it

PR #155, opened 2023-09-27
Allow for precomputed quantum kernels in quantum kernel methods

JanSchnabel:
* Solves [Issue#144](https://github.com/sQUlearn/squlearn/issues/144)
* After consulting @FredericRapp decided to leave QGP methods as is

### DIFF
--- a/src/squlearn/kernel/ml/qgpr.py
+++ b/src/squlearn/kernel/ml/qgpr.py
@@ -5,6 +5,7 @@ from ..matrix.kernel_matrix_base import KernelMatrixBase
 from ..matrix.regularization import regularize_full_kernel
 
 import numpy as np
+from typing import Optional, Union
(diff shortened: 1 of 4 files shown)


rothma:
Why neglect QGP?

JanSchnabel:
> Why neglect QGP?

Since sklearn also neglects it :D This is a weak argument, I know. Thus, I can work on it and also allow for the "precomputed" option in QGPs

JanSchnabel:
@FredericRapp , can you please check QGPR?

Since QGPC wraps to the respective sklearn method, which does not provide a "precomputed" option, I decided to neglect this

FredericRapp (approved):
Changes in QGPR look good, and the tricky part with the indices of the K_testtrain seems to be correct.

MoritzWillmann (approved):
Looks good

Issue #160, opened 2023-09-28
Remove output while training QNNs


MoritzWillmann:
Remove this
```
theta [0.77770241 0.23754122 0.82427853 0.9657492  0.97260111 0.45344925
 0.60904246 0.77552651 0.64161334 0.72201823 0.03503652 0.29844947
 0.05851249 0.85706094 0.37285403 0.67984795 0.25627995 0.34758122
 0.00941277]
Current value: 15.064846438837378
theta [0.77770241 0.23754122 0.82427853 0.9657492  0.97260111 0.45344925
 0.60904246 0.77552651 0.64161334 0.72201823 0.03503652 0.29844947
 0.05851249 0.85706094 0.37285403 0.67984795 0.25627995 0.34758122
 0.00941277]
Grad length: 23.534559613237978
grad [ -0.05716598  -1.1969308   -2.5070993   -0.15684358  -0.26523052
  -0.92167099  -1.99965822  -0.11460244   3.47621286  -3.85943808
  -1.36884346  -0.15684358   2.57708562  -3.34091146  -0.90915976
  -0.11460244  19.64618798  -1.28281638 -10.29850044]
```

or at least move it to a logger

PR #166, opened 2023-10-04
Draft: Clean qnn output

MoritzWillmann:
This PR:
- removes all prints in qnns fit functions
- adds an option to add callback to optimizers (either in `__init__` or via `set_callback`)
- adds an option to specify a callback in QNN high level methods
- demonstrates how to use tqdm progress bar in the QNN examples
- solves #160 

I still need to:
- maybe test this
- use in all examples


David-Kreplin:
I implemented a callback for Adam and adjusted the wrapper for SPSA. 
What is left is a godd default, so we that we have a progressbar as default.

David-Kreplin (approved):

PR #175, opened 2023-10-13
Automated nb tests

MoritzWillmann:
This PR adds notebook tests to the test workflow


David-Kreplin (approved):

MoritzWillmann:
The workflow runs very long, maybe we need to discuss on tuesday if we really want this

PR #188, opened 2023-10-30
Fix number of shots in executor (resolves testjobs errors)

David-Kreplin:


src/squlearn/util/executor.py
```diff
@@ -849,14 +852,15 @@ def get_shots(self) -> int:
                         "The number of shots of the given \
                                       Estimator and Sampler is not equal!"
                     )
-
-            shots = max(shots_estimator, shots_sampler)
-
         elif self._backend is not None:
             shots = self._backend.options.shots
+            if "statevector_simulator" in str(self._backend):
+                shots = 0
```
rothma, line 858: Why set it to 0 if you set shots to None when they are zero just a few lines afterwards?

rothma (approved):

PR #189, opened 2023-10-31
Harmonize use of X_train, X_test, y_train as arguments in fit() and predict() with sklearn to X and y

JanSchnabel:


JanSchnabel:
> Looks good, but there is an additional qgpr_model.fit(...) and .predict(...) in the BayesOpt tutorial which needs to be adapted.

Should be fixed now.

FredericRapp (approved):

PR #195, opened 2023-11-08
Mlflow integration

MoritzWillmann:
This PR:
- enables the use of sQUlearn with MLFlow in the sklearn flavor
- adds an example that shows how to do so
- enforces the use of the local `qiskit_settings.conf` when building the docs locally


David-Kreplin (approved):
Very nice! Thank you.

PR #197, opened 2023-11-10
Multiple fixes for the release

David-Kreplin:
- Fixes missing encoding circuit in kernel matrix parameters
- Adds seed for Executor 
- Caching Option for the ProjectedQuantumKernel
- Bug fix in QNNs that seeded qasm results are ending in different solutions
- Bug fix for Shot Control
- Bug fix for setting shots in the Executor
- Test for setting shots and getting primitives of the Executor


MoritzWillmann:
This massively elongates the runtime of the `qgpc_workflow.ipynb` workflow. This is also why the test workflow fails.

MoritzWillmann:
> This massively elongates the runtime of the `qgpc_workflow.ipynb` workflow. This is also why the test workflow fails.

The problem seems to be withing predict_proba (but also fit runs fairly long)

MoritzWillmann:
The issue concerning the implementation of fidelity quantum kernels is noted in issue #198

MoritzWillmann (approved):

PR #208, opened 2023-11-16
bug fix for training without observable parameters

David-Kreplin:


MoritzWillmann (approved):

MoritzWillmann:
The error seems to be on scipy's documentation website so we can still merge this 😄

PR #213, opened 2023-11-27
use `FidelityStatevectorKernel`

### LINKED ISSUE #198: `qiskit_machine_learning.kernels.QuantumKernel` is deprecated as of `0.7.0`
We need to find an alternative. Sadly the implemenation of `qiskit_machine_learning.kernels.FidelityQuantumKernel` (which is also used by `qiskit_machine_learning.kernels.TrainableFidelityQuantumKernel` internally) is very slow and thus not an alternative at the moment.

MoritzWillmann:
This PR:
- will introduce the use of `FidelityStatevectorKernel` and `TrainableFidelityStatevectorKernel` in `FidelityKernel`.
- makes `FidelityKernel` use the Statevector kernels, if the `Executor._backend` contains the string `"statevector_simulator"`, else the `FidelityQuantumKernel` based implementations will be used.
- will bump the required version of `qiskit-machine-learning` to `>=0.6.1`
- resolves #198


David-Kreplin (approved):
Looks great, thank you!

PR #217, opened 2023-12-01
Add qcnn encoding circuit with example.

MHagelueken:
The qcnn encoding circuit is added and an example which is supposed to show how to work with it placed in the qnn folder.


David-Kreplin:
Thanks for the update, unfortunately the example you created exceeds the time constraints we have for testing it. 
Could you find a quicker example, maybe initialize with suitable parameters first. 
Also you need to fix some formatting which you can do automatically with a tool called black. Call me, then I can show you how it is done.

rothma:
For discussion: Do we want to emphasize the QCNN by also adding a high-level QCNN class which alread works as a QCNN out-of-the-box or are we fine with the user having to build a QCNN via QNN + QCNN Encoding circuit as it is now?

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,577 @@
...
+                            break
+
+    def set_num_features(self, n: int = 0):
+        """Change the number of features in the feature map of the circuit."""
+        if n > 0:
+            self._num_features = n
+
+    def convolution(
+        self,
+        QC="None",
```
MoritzWillmann, line 87:
```suggestion
        quantum_circuit=None,
```
for style purposes, let's change `QC` to `quantum_circuit`. This also applies to all later occurences of `QC`. Also change `"None"` to `None`

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,577 @@
...
+        Args:
+            QC Union[EncodingCircuitBase,QuantumCircuit]: The quantum circuit, which is applied on every qubit
+                modulo qubits of this circuit.
+            label (str): Sets the name of the operation.
+            alternating (bool): It applies the QC on every qubit modulo qubits of this circuit beginning at 0
+                and if True it applies the QC on every qubit beginning at 1 again.
+            diff_params (bool): If True, different parameters are used for the gates build by this layer.
+        """
+        # Define default circuit
+        if QC == "None":
```
MoritzWillmann, line 105:
```suggestion
        if not quantum_circuit:
```
Using `None` also allows for a more readable if statement here

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,577 @@
...
+
+    def set_num_features(self, n: int = 0):
+        """Change the number of features in the feature map of the circuit."""
+        if n > 0:
+            self._num_features = n
+
+    def convolution(
+        self,
+        QC="None",
+        label: str = "C",
```
MoritzWillmann, line 88: We could have enough space to write `Conv`, `Pool` and `FC`. This could make stuff more readable, but I'm not hard-set on this

MoritzWillmann (commented):
In general looks good, Just some style requests

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,617 @@
...
+
+from squlearn.encoding_circuit.encoding_circuit_base import EncodingCircuitBase
+from squlearn.encoding_circuit.circuit_library.param_z_feature_map import ParamZFeatureMap
+from squlearn.observables import CustomObservable
+
+
+class QCNNEncodingCircuit(EncodingCircuitBase):
+
+    """
+    Instantiate the object for building a QCNN.
```
MoritzWillmann, line 15:
```suggestion
    Encoding circuit for quantum convolutional neural networks (QCNN).
    
```

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,617 @@
...
+    One can hand over the optional arguments "mpl" to get the circuit in matplotlib drawing and "decompose = True"
+    to resolve the individual components of the circuit building blocks
+    (e.g. mycircuit.draw("mpl", decompose = True)).
+
+    Args:
+        num_qubits (int): Number of qubits of the QCNN encoding circuit.
+        num_features (int): Dimension of the feature vector. By default this is 0,
+            so an feature map must be provided (or can be changed to x by set_num_features(x)).
+            If num_features is bigger then 0, then in the get_circuit function a ZFeatureMap is build.
+        default (bool): If True, the default circuit is build.
```
MoritzWillmann, line 26: "used" or "built", not "build"

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,617 @@
...
+
+
+class QCNNEncodingCircuit(EncodingCircuitBase):
+
+    """
+    Instantiate the object for building a QCNN.
+    In order to plot your circuit (called e.g. mycircuit) at any given point use: mycircuit.draw().
+    One can hand over the optional arguments "mpl" to get the circuit in matplotlib drawing and "decompose = True"
+    to resolve the individual components of the circuit building blocks
+    (e.g. mycircuit.draw("mpl", decompose = True)).
```
MoritzWillmann, line 19: This description is a bit unnecessary. I think this section should rather be used to describe the Idea of QCNNs. Also make references to the paper(s) that introduced the concept.

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,617 @@
...
+        if len(self.left_qubits) <= 1:
+            print("The circuit has too few qubits")
+            return
+        self.convolution()
+        self.pooling()
+        if len(self.left_qubits) > 1:
+            self.repeat_layers()
+        self.fully_connected()
+
+    def QCNNObservable(self, pauli: str = "Z"):
```
MoritzWillmann, line 553: Add return type

```suggestion
    def QCNNObservable(self, pauli: str = "Z") -> CustomObservable:
```

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,617 @@
...
+                if i == j:
+                    observable = pauli + observable  # most right qubit in operator refers to q0
+                else:
+                    observable = "I" + observable
+            observable_list.append(observable)
+        return CustomObservable(
+            num_qubits=self.num_qubits, operator_string=observable_list, parameterized=True
+        )
+
+    def convert_encoding_circuit(self, quantum_circuit):
```
MoritzWillmann, line 577:
```suggestion
    def convert_encoding_circuit(self, quantum_circuit) -> QuantumCircuit:
```

MoritzWillmann (commented):
Okay, there's more...
- Please check the Docstrings so that they match the style guide. For more information see [the docstring section in our contribution guidelines](https://github.com/sQUlearn/squlearn/blob/main/.github/CONTRIBUTING.md#docstrings).
- Some functions seem to be fully internal (they shouldn't be called from the outside). Please make them either protected or private (check [this](https://www.tutorialsteacher.com/python/public-private-protected-modifiers) for more information)
- Add return type hints to the functions that return something

MoritzWillmann:
> For discussion: Do we want to emphasize the QCNN by also adding a high-level QCNN class which alread works as a QCNN out-of-the-box or are we fine with the user having to build a QCNN via QNN + QCNN Encoding circuit as it is now?

no

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,686 @@
...
+    def repeat_layers(self, n_times: int = 0):
+        """
+        Repeat the already applied gates to simply build the circuit.
+
+        This does not work with a pooling layer with supplied in- and output lists.
+
+        Args:
+        n_times (int): The number of times the already applied gates are repeatedly applied.
+        default configuration: At least once applied and until less then 4 qubits are left
+            and only once if there is no pooling gate applied.
```
MoritzWillmann, line 563: Indents
```suggestion
            n_times (int): The number of times the already applied gates are repeatedly applied.
            default configuration: At least once applied and until less then 4 qubits are left
                and only once if there is no pooling gate applied.
```

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,686 @@
...
+
+    def QCNNObservable(self, pauli: str = "Z") -> CustomObservable:
+        """
+        Build a fitting observable for the current circuit.
+
+        This function should be called after beeing finished building the circuit.
+        It only contains single qubit measurements.
+
+        Args:
+        pauli (str): Its the used pauli gate so either X,Y or Z.
```
MoritzWillmann, line 619: Indent
```suggestion
            pauli (str): Its the used pauli gate so either X,Y or Z.
```

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,686 @@
...
+        """
+        Build the circuit "backwards".
+
+        Build the circuit by supplying the number of qubits which should be left
+        after the already supplied gates. This function then generates the necessary number
+        of initial qubits and applies the supplied gates.
+
+        Args:
+        final_num_qubits (int):
+            The number of qubits which should be left after applying the supplied gates.
```
MoritzWillmann, line 661: Indent
```suggestion
            final_num_qubits (int):
                The number of qubits which should be left after applying the supplied gates.
```

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,686 @@
...
+        """
+        Add a fully connected layer to the encoding circuit.
+
+        The fully connected layer should be placed at the end
+        and operates on all qubits remaining in the circuit.
+
+        Args:
+        quantum_circuit Union[EncodingCircuitBase,QuantumCircuit, None]:
+            The quantum circuit, which is applied in this layer.
+        label: The name of the layer.
```
MoritzWillmann, line 356: Indents
```suggestion
            quantum_circuit Union[EncodingCircuitBase,QuantumCircuit, None]:
                The quantum circuit, which is applied in this layer.
            label: The name of the layer.
```

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,686 @@
...
+            within the list must be equal to the number of qubits of input circuit.
+        output_list (list): Exactly if an input list is entered, an output list must be entered.
+            The output list defines the qubits which are left in the circuit to operate on.
+            The list should be structured as: [[qubit1,qubit2,..],[qubit3,qubit4,..],..].
+            It must have the same length as the input list and in each sublist
+            its elements must be in the corresponding input sublist
+            while beeing at least one element less.
+        THE QUBIT NUMBERS IN THE LISTS REFER TO THE INITIAL QUBIT NUMBERS!
+        Default circuit: Entangles qubit i and qubit i+1.
+            Only qubit i stays in the circuit for further operations.
```
MoritzWillmann, line 204: What is this? It seems like this should be in the description not in the argument list.

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,686 @@
...
+            The input list defines the qubits the input circuit acts on.
+            The list should be structured as: [[qubit1,qubit2,..],[qubit3,qubit4,..],..].
+            Every qubit can only be adressed once and the number of qubits in each list
+            within the list must be equal to the number of qubits of input circuit.
+        output_list (list): Exactly if an input list is entered, an output list must be entered.
+            The output list defines the qubits which are left in the circuit to operate on.
+            The list should be structured as: [[qubit1,qubit2,..],[qubit3,qubit4,..],..].
+            It must have the same length as the input list and in each sublist
+            its elements must be in the corresponding input sublist
+            while beeing at least one element less.
```
MoritzWillmann, line 201: Indents. Also description for `_new_operation` is missing

```suggestion
            quantum_circuit Union[EncodingCircuitBase,QuantumCircuit, None]:
                The quantum circuit, which is applied in this layer.
                Must be an entangling layer, which entangles qubits.
            label (str): The name of the layer.
            measurement (bool): Sets whether the qubits,
                which are not used anymore after this layer, are measured.
                If True, quantum_circuit must consist of exactly one classical bit
                additionally to the quantum bits.
            input_list (list): Optionally one can pass the structure of the gates operating.
                The input list defines the qubits the input circuit acts on.
                The list should be structured as: [[qubit1,qubit2,..],[qubit3,qubit4,..],..].
                Every qubit can only be adressed once and the number of qubits in each list
                within the list must be equal to the number of qubits of input circuit.
            output_list (list): Exactly if an input list is entered, an output list must be entered.
                The output list defines the qubits which are left in the circuit to operate on.
                The list should be structured as: [[qubit1,qubit2,..],[qubit3,qubit4,..],..].
                It must have the same length as the input list and in each sublist
                its elements must be in the corresponding input sublist
                while beeing at least one element less.
```

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,686 @@
...
+        if n > 0:
+            self._num_features = n
+
+    def convolution(
+        self,
+        quantum_circuit: Union[QuantumCircuit, EncodingCircuitBase, None] = None,
+        label: str = "Conv",
+        alternating: bool = True,
+        diff_params: bool = True,
+        _new_operation: bool = True,
```
MoritzWillmann, line 106: Docstring description for `_new_operation` is missing. This also applies for other functions.

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,686 @@
...
+    "Quantum convolutional neural networks" by Cong, Iris and Choi, Soonwon and Lukin, Mikhail D.
+
+    Args:
+        num_qubits (int): Number of initial qubits of the QCNN encoding circuit.
+        num_features (int): Dimension of the feature vector.
+            By default this is 0, so a feature map must be provided.
+            If the number of features is bigger then 0,
+            then in the get_circuit function a ZFeatureMap is built to encode the features.
+        default (bool): If True, the default circuit is built.
+    """
```
MoritzWillmann, line 29: A bit clearer and right way to reference
```suggestion
    The structure is inspired by classical convolutional neural networks. The number of active
    qubits reduces with each layer. The design idea was initially proposed in reference [1].

    Args:
        num_qubits (int): Number of initial qubits of the QCNN encoding circuit.
        num_features (int): Dimension of the feature vector.
            By default this is 0, so a feature map must be provided.
            If the number of features is bigger then 0,
            then in the get_circuit function a ZFeatureMap is built to encode the features.
        default (bool): If True, the default circuit is built.
    
    References
    -----------
    [1]: `Cong, I., Choi, S. & Lukin, M.D. Quantum convolutional neural networks. Nat. Phys. 15,
    1273–1278 (2019). <https://doi.org/10.1038/s41567-019-0648-8>`_
    """
```

David-Kreplin:
I changed this to a draft, before it gets accidentally merged.

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,718 @@
...
+                Every qubit can only be adressed once and the number of qubits in each list
+                within the list must be equal to the number of qubits of input circuit.
+            output_list (list):
+                Exactly if an input list is entered, an output list must be entered.
+                The output list defines the qubits which are left in the circuit to operate on.
+                The list should be structured as: [[qubit1,qubit2,..],[qubit3,qubit4,..],..].
+                It must have the same length as the input list and in each sublist
+                its elements must be in the corresponding input sublist
+                while beeing at least one element less.
+            THE QUBIT NUMBERS IN THE LISTS REFER TO THE INITIAL QUBIT NUMBERS!
```
MoritzWillmann, line 214:
```suggestion
                THE QUBIT NUMBERS IN THE LISTS REFER TO THE INITIAL QUBIT NUMBERS!
```

src/squlearn/encoding_circuit/qcnn_encoding_circuit.py
MoritzWillmann: The reference section in the bottom is still missing

MoritzWillmann (approved):

src/squlearn/encoding_circuit/circuit_library/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,721 @@
...
+
+    def set_num_features(self, n: int = 0):
+        """
+        Change the number of features in the feature map of the circuit.
+
+        Args:
+            n (int): The new number of features.
+        """
+        if n > 0:
+            self._num_features = n
```
MoritzWillmann, line 100: `squlearn.encoding_circuit.EncodingCircuitBase` has `set_params` implemented. I'd suggest overwriting and using this function.

src/squlearn/encoding_circuit/circuit_library/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,708 @@
...
+        this will throw a error in the troubling layer.
+
+        Args:
+            params: Hyper-parameters (num_qubits or num_features)
+                and their values, e.g. ``num_qubits=2``.
+        """
+        super().set_params(**params)
+        if "num_qubits" in params:
+            self._left_qubits = [i for i in range(self.num_qubits)]
+            self._num_parameters = 0
```
MoritzWillmann, line 75: now, if we set both `num_qubits` and `num_params`, `num_params` will always be `0`. Is this wanted behavior?

MHagelueken: The number of parameters of the encoding circuit are determined automatically by the gates applied and therefore should not be set externally. The set_params function should not have a keyword "num_parameters". Therefore this should be appropriate, since when the number of qubits change it must be newly calculated how many parameters are needed to parametrize the given circuit.

MoritzWillmann: ah my bad, I mixed it up with `num_features`. You're totally right :)

src/squlearn/encoding_circuit/circuit_library/qcnn_encoding_circuit.py
```diff
@@ -0,0 +1,709 @@
...
+                        (j + 1) * quantum_circuit.num_qubits + 1,
+                    ):
+                        i_1 = i
+                        if i == len(left_qubits):  # sothat the last wraps with the first qubit
+                            i_1 = 0
+                        adressed_qubits.append(left_qubits[i_1])
+                    total_qc = total_qc.compose(
+                        circuit_to_gate(quantum_circuit),
+                        qubits=adressed_qubits,
+                    )
```
MoritzWillmann, line 500: If i can ever be bigger than `len(left_qubits)`
```suggestion
                    total_qc = total_qc.compose(
                        circuit_to_gate(quantum_circuit),
                        qubits=[
                            left_qubits[i] if i != len(left_qubits) else left_qubits[0]
                            for i in range(
                                j * quantum_circuit.num_qubits + 1,
                                (j + 1) * quantum_circuit.num_qubits + 1,
                            )
                        ],
                    )
```
else one could even do
```suggestion
                    total_qc = total_qc.compose(
                        circuit_to_gate(quantum_circuit),
                        qubits=[
                            left_qubits[i % len(left_qubits)]
                            for i in range(
                                j * quantum_circuit.num_qubits + 1,
                                (j + 1) * quantum_circuit.num_qubits + 1,
                            )
                        ],
                    )
```

David-Kreplin (approved):

PR #222, opened 2024-01-11
Add feature loop

MoritzWillmann:
This PR allows all encoding circuits to utilize more features than qubits. It also changes the name of some variables to be a bit more meaningful.


David-Kreplin:
There seems to be a bug that was introduced by this PR. Another PR has gone through, so it is certainly this PR.

David-Kreplin (approved):

PR #224, opened 2024-01-15
Adjust inital parameters for chebyshev encoding circuits

David-Kreplin:


src/squlearn/encoding_circuit/circuit_library/chebyshev_pqc.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 from typing import Union
+import math
```
MoritzWillmann, line 3: I would suggest using `np.ceil` and omitting this import just for cleanliness

David-Kreplin:
Math imports are removed

MoritzWillmann (approved):
Noice

PR #227, opened 2024-01-18
Add qcnn encoding circuit

MHagelueken:


David-Kreplin:
For checking if the backend is a statevector simulator, please use:
if "statevector_simulator" in str(executor.backend)
That captures also IBMQ statevector simulators or AER statevector_simulators.
Maybe we should add executor.statevector_backend

David-Kreplin:
I will add a functionality to swap to the BackendPrimtives automatically

David-Kreplin (changes requested):
Look like the function _build_reorder is not needed and can be removed?

Another hopefully small thing is the routine evaluate_tree_with_sampler (Line 1403 ff).
This routine is not called in sQUlearn at the moment, so it is not super crutial.

During this routine, there is some measurment added, if nothing exist (Line 1467 ff):

```
            total_circuit_list += [
                circuit.measure_all(inplace=False) if circuit.num_clbits == 0 else circuit
                for circuit in circuit_list
            ]
```

Could you pleace compare this with your replacement code in the evaluate_with_sampler routine. In case it is not much work, could you fix it?

David-Kreplin (approved):
Finally done :)

PR #229, opened 2024-02-02
Change order in QNN evaluation routine 

David-Kreplin:


examples/qnn/regression_example.ipynb
MoritzWillmann, line 1: The `pqc.draw("mpl")` call throws a warning. Can you please update your local default style to not throw this warning 🙂

MoritzWillmann:
Is there any apparent reason to change the order? Just asking out of curiosity

David-Kreplin:
As discussed, the previous interface required an obscure tuple construction which is avoided by the new order and utilizing *args

MoritzWillmann (approved):

PR #236, opened 2024-02-15
bug fix for evaluations with multiple parameters

David-Kreplin:


src/squlearn/qnn/qnn.py
```diff
@@ -938,55 +938,67 @@ def flatten(container):
...
-
-                # If there are lists in the operators, add them next (e.g. dfdop)
-                if len(ilist) > length:
-                    if self.multiple_output:
-                        swapp_list = swapp_list + ilist[3 + num_nested : -1]
-                    else:
-                        swapp_list = swapp_list + ilist[3 + num_nested :]
-
-                # If there are lists in the circuits, add them here (e.g. dfdp)
+                swapp_list2 = [ilist[2 + num_nested]]
```
MoritzWillmann, line 968: Is there any specific reason why this is called `swapp_list2` and later `swapp_list3`? IMO we could just call it `swap_list` everywhere. Also `ilist` would be better called `index_list` or `indices`.

David-Kreplin:
Updated variable names following Moritz's suggestions.

MoritzWillmann (approved):

PR #237, opened 2024-02-19
Upgrade to Qiskit 1.0.0

MoritzWillmann:
This PR incorporates fixes needed to use the new Qiskit 1.0.0. This includes:
- using the standalone `qiskit-algorithms` package
- using the standalone `qiskit-aer` package everywhere, where it wasn't used yet
- using the `aer_simulator` instead of `statevector_simulator` and `qasm_simulator` with according settings

So far the tests still fail, due to `qiskit-algorithms` not being updated to work with Qiskit 1.0.0 yet. I will expect an update soon and will then fix the version and set the PR to ready for review.
- [x] fix `qiskit-algorithm` version

## Note
If you want to use squlearn until the update, be sure to install Qiskit<1.0.0


src/squlearn/util/executor.py
```diff
@@ -933,7 +937,7 @@ def get_shots(self) -> int:
 
             shots = max(shots_estimator, shots_sampler)
         elif self._backend is not None:
-            if "statevector_simulator" not in str(self._backend):
+            if "statevector" not in self._backend.configuration().backend_name:
```
David-Kreplin, line 940: Did you try this with real backends? 
Sometimes backend.configuration() does not exist

MoritzWillmann: Works with each of the freely available backends, didn't check for paid backends

David-Kreplin: If works for the free ones, its fine. Unfortunatly this change occurs in a couple of other places as well :(

David-Kreplin: Sorry, see my review for an failing example :D

MoritzWillmann:
You're right i implemented `backend_name` and `is_statevector` functions to handle this

src/squlearn/kernel/matrix/fidelity_kernel.py
```diff
@@ -101,8 +101,7 @@ def __init__(
         self._enc_circ = self._encoding_circuit.get_circuit(
             self._feature_vector, self._parameter_vector
         )
-
-        if "statevector_simulator" in str(self._executor._backend):
+        if "statevector" in self._executor._backend.configuration().backend_name:
```
David-Kreplin, line 104: is_statevector is missing here

David-Kreplin (changes requested):
is_statevector is missing in fidelity_kernel.py (Line 104)

src/squlearn/kernel/matrix/fidelity_kernel.py
MoritzWillmann: You're right, thanks for the pointer 😄 
```suggestion
        if self._executor.is_statevector:
```

David-Kreplin (approved):
Thank you very much for your effort!

PR #242, opened 2024-04-09
Add SGLBO optimizer

DennisKleinhans:
- add file sglbo.py => `SGLBO` optimizer 
- extended *\_\_init\_\_.py* with the `SGLBO` class 
- extended _classes.rst_ with `optimizers.SGLBO` to build the docs for the optimizer 
- add tests for the `SGLBO` class

### DIFF
--- a/docs/modules/classes.rst
+++ b/docs/modules/classes.rst
@@ -229,6 +229,7 @@ Implemented optimizers
    optimizers.LBFGSB
    optimizers.SLSQP
    optimizers.SPSA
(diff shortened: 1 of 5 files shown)


MoritzWillmann:
Overall looks good, we need an implementation for `_update_lr` though. I'd say just a simple pass will be okay in this case

MoritzWillmann:
Also a test case that utilizes the `step` function would be nice

MoritzWillmann (approved):
looks good to me

PR #253, opened 2024-04-22
correct example for the qcnn encoding circuit

MHagelueken:
Change a few lines of code and correct descriptions.


David-Kreplin:
Title says that this is a classification example, but the one given is a regression example. 
We could also show the example of the paper, which is a classification example.

David-Kreplin (approved):

Issue #271, opened 2024-06-06
Second order (and mixed) QNN Pennylane derivatives with shots are not working ( dfdxdp, dfdxdx, dfdxdop)


rupof:
**Describe the bug**

Simulations with high order QNN Pennylane derivatives with shots (dfdxdp, dfdxdx, dfdxdop) give different results when comparing to qiskit_shots, qiskit_statevector and penylane_statevector (all giving the correct results).

**To Reproduce**

I calculated f, dfdx, dfdxdp, dfdxdx, for a simple qnn with ChebyshevRx encoding:

```
import numpy as np

from squlearn.observables import SummedPaulis
from squlearn.encoding_circuit import QiskitEncodingCircuit, ChebyshevRx
from squlearn.qnn.lowlevel_qnn import LowLevelQNN
from squlearn import Executor



num_qubits = 2
x_array = np.array([[0.75], 
                    [0.1]])

circuit = ChebyshevRx(num_features=1, num_qubits=num_qubits, num_layers=1)
observable = SummedPaulis(num_qubits)

qnn_pennylane_shots = LowLevelQNN(circuit, observable, Executor("pennylane", shots=50000))
qnn_pennylane_statevector = LowLevelQNN(circuit, observable, Executor("pennylane"))

qnn_qiskit_shots = LowLevelQNN(circuit, observable, Executor("qiskit", shots=50000))
qnn_qiskit = LowLevelQNN(circuit, observable, Executor("qiskit"))

np.random.seed(1)
param = np.random.rand(qnn_pennylane_shots.num_parameters)
param_obs = np.random.rand(observable.num_parameters)


print("Pennylane statevector ")
print("f\n", qnn_pennylane_statevector.evaluate(x_array, param, param_obs, "f")["f"])
print("dfdx\n", qnn_pennylane_statevector.evaluate(x_array, param, param_obs, "dfdx")["dfdx"])
print("dfdpdx\n", qnn_pennylane_statevector.evaluate(x_array, param, param_obs, "dfdpdx")["dfdpdx"])
print("dfdxdx\n", qnn_pennylane_statevector.evaluate(x_array, param, param_obs, "dfdxdx")["dfdxdx"])
#print("dfdxdop\n", qnn_pennylane_statevector.evaluate(x_array, param, param_obs, "dfdopdx")["dfdopdx"])


print("-----------------")
print("Pennylane shots") 
[...]

PR #272, opened 2024-06-07
fix for issue #271

David-Kreplin:
fix for issue #271


David-Kreplin:
There is an issue with the dxdx derivative and the arcos function, that has to be fixed before merging.

MoritzWillmann:
Will we need to change things back when things are fixed in Pennylane?

David-Kreplin:
I think, no further adjustments are necessary. But we have to wait until the PennyLane issue is fixed to be sure.

MoritzWillmann (approved):

PR #276, opened 2024-06-14
Replacement of Space with Bounds for the definition of the search space

### LINKED ISSUE #269: Another skopt dependency (in the examples)
**Describe the bug**
Similar to #267, we have a dependency to skopt that crashes our test. This one is in `examples/tutorials/quantum_bayesian_optimization.ipynb`. We need to find another solution here aswell.

**To Reproduce**
Run the stated example with a sklearn version >=1.5.0

DennisKleinhans:
No longer use of  `Space` from the `skopt.Space` module. Instead `Bounds` from the `scipy.optimize` module is used to define the search space.

closes #269


David-Kreplin:
Please run black through the notebooks (black --ipynb examples)

MoritzWillmann (approved):

PR #296, opened 2024-10-09
fix docs

MoritzWillmann:
- fix `sphinx_rtd_theme` to version `<3.0.0`
- make all strings with escape sequences raw
- clean some dependencies for dev, docs and examples


David-Kreplin (approved):
Thanks!

Issue #292, opened 2024-08-19
Handle feature encoding when #features != n*encoding slots


rothma:
- Check that all encoding circuits behave similar (if not eplicitly different like high-dim)
- Desired behaviour: fill from top to bottom with "period boundary conditions" (also across layers)

PR #301, opened 2024-10-25
Remove need for `num_features` before fitting 

### LINKED ISSUE #266: Remove need for `num_features` before fitting
This came up recently while discussing the integration of sQUlearn into the AutoQML lib. Basically we need to remove any references to `num_features` in the standard workflow for the high level methods.

## Status Quo
Currently we require `num_features` in the initialization of the encoding circuit which is itself required in the creation of the high level method.
```python
encoding_circuit = EncodingCircuit(
    ...,
    num_features=...
    ...
)
...
clf = QuantumClassifier(
    ...,
    encoding_circuit=encoding_circuit,
    ...
)
```

## Desired
This is not really good etiquette and we want to remove the need to know `num_features` in the instantiation of the high level method. Instead we want to use the dimension that is implicitly provided by `X` in the call to the `fit` method.

DennisKleinhans:
### 1. Delayed Initialization of High-Level Methods
- The `num_features` parameter is no longer required during the instantiation of high-level methods. 
- If `num_features` is not provided, the complete initialization of the high-level method is postponed until the `fit` method is called, allowing the actual number of features to be known.
- If `num_features` is provided during instantiation, the high-level method will be fully initialized immediately.

### 2. New `num_encoding_slots` Property for `EncodingCircuit`
- A new property called `num_encoding_slots` has been introduced for each `EncodingCircuit`, which calculates the number of encoding slots individually.
- In the `get_circuit` method, there is now a check to ensure that the number of provided features matches the number of available encoding slots. 
- If there are not enough encoding slots for all features, an `EncodingSlotsMismatchError` is raised.

closes #266, #292


MoritzWillmann:
Hey @DennisKleinhans thanks for your contribution! Please make sure, that all tests, notebooks and docs work 😄 
Cheers Moritz

David-Kreplin:
Please check random_encoding_circuit.py, random_layered_encoding_circuit.py, pruned_encoding_circuit.py and transpiled_encpding_circuit.py

tests/qnn/test_qnnr.py
```diff
@@ -57,6 +57,7 @@ def test_predict_unfitted(self, qnn_regressor, data):
             - whether the prediction output is correct
         """
         X, y = data
+        qnn_regressor._initialize_lowlevel_qnn()
```
MoritzWillmann, line 60: recheck with desired behavior

tests/qnn/test_training.py
MoritzWillmann, line 1: make features unset as the default, you can add some tests where it is set, though

MoritzWillmann (changes requested):
Generally it is good. We need to change some stuff though:
- it should be only possible to set `num_features` for backwards-compatibility reasons, nothing should be done with it in `__init__` except for storing it in case of the encoding circuits (hlms shouldn't store it at all)
- the encoding circuit's `get_circuit` should do all the setups and they should use the dimension of `X` instead of `num_features`
- there should only be a check for consistency in the `get_circuit` of the encoding circuits if `num_features` was set

Please adapt this accordingly and also adapt the tests to check for this desired behavior. Sometimes i only commented sparsely because I didn't want to comment the same stuff multiple times. If there are any questions, please don't hesitate to ask 🙂 

Cheers

src/squlearn/kernel/qgpc.py
```diff
@@ -164,10 +163,10 @@ def quantum_kernel(self, quantum_kernel: KernelMatrixBase):
         self._quantum_kernel = quantum_kernel
         self.kernel = kernel_wrapper(quantum_kernel)
 
-    def __initialize(self) -> None:
+    def __initialize(self, num_features: int) -> None:
```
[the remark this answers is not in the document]
MoritzWillmann: This also holds for all other Kernel HLMs

MoritzWillmann (changes requested):

src/squlearn/encoding_circuit/encoding_circuit_base.py
```diff
@@ -290,28 +298,17 @@ def parameter_bounds(self) -> np.ndarray:
 
             @property
             def feature_bounds(self) -> np.ndarray:
-                """Returns the bounds of the features of composed encoding circuit.
+                """Returns the bounds of the features of composed encoding circuit. To get the bounds for a specific number of features, use get_feature_bounds().
```
MoritzWillmann, line 301:
```suggestion
                """
                Returns the bounds of the features of composed encoding circuit.
                
                To get the bounds for a specific number of features, use get_feature_bounds().
```

src/squlearn/encoding_circuit/encoding_circuit_base.py
```diff
@@ -394,6 +391,17 @@ def set_params(self, **params) -> None:
                 if len(ec2_dict) > 0:
                     self.ec2.set_params(**ec2_dict)
 
+            def set_num_features(self, num_features_ec1: int, num_features_ec2: int) -> None:
```
MoritzWillmann, line 394: do we need a setter?

tests/encoding_circuit/test_encoding_circuit_base.py
```diff
@@ -71,13 +85,26 @@ def test_add(self):
...
-            circuit.set_params(invalid_param=1)
+            composed_circuit.get_circuit([], [])
+
+        composed_circuit.set_num_features(2, 2)
+
+        expected_circuit = QuantumCircuit(2)
+        expected_circuit.h(range(2))
+        expected_circuit.h(range(2))
+
+        assert str(composed_circuit.get_circuit([], [])) == str(expected_circuit)
```
MoritzWillmann, line 106: this doesn't make a lot of sense

src/squlearn/qnn/lowlevel_qnn/lowlevel_qnn_pennylane.py
```diff
@@ -111,6 +113,7 @@ def __init__(
         parameterized_quantum_circuit: EncodingCircuitBase,
         observable: Union[ObservableBase, list],
         executor: Executor,
+        num_features: int,
```
MoritzWillmann, line 116: do we need this here?

src/squlearn/encoding_circuit/circuit_library/random_layered_encoding_circuit.py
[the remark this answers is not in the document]
DennisKleinhans: Could you please specify what behaviour you mean? I don't understand what changes you want...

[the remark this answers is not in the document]
MoritzWillmann: In `RandomEncodingCircuit` we only generate a random configuration once, in `RandomLayeredEncodingCircuit` a configuration is generated each time `get_circuit` is called.

MoritzWillmann (approved):
Looks good now, LGTM

PR #307, opened 2024-12-10
Tutorial for the ODE QNN Regressor 

rupof:
Hi! 

In this PR, I made a tutorial for the QNN ODE solver. 

The [logistic differential equation](https://en.wikipedia.org/wiki/Logistic_function), a special case of the [Bernoulli Differential Equation](https://en.wikipedia.org/wiki/Bernoulli_differential_equation), is solved and compared to its analytical solution.  

If this example is too trivial, I can search for another one. 

To include the tutorial, I also:
	- Created `docs/examples/example_quantum_bayesian_optimization.nblink` (which I think it was necessary but I am not sure)
	- Included `--ignore=examples/tutorials/qnn_ode_solver.ipynb`, in `pyproject.toml` so that it is not checked by pychecks (as the solution takes a long time)

Furthermore:
	- I fixed one small typo in `examples/tutorials/quantum_bayesian_optimization.ipynb`, "structre->structure"
	- I fixed a small bug in the `KyriienkoEncodingCircuit` for the `chebyshev_product`

Feel free to let me know what should I improve or change. Thankss!

Roberto


David-Kreplin:
After our conversiation, here the Todos:

- [ ] setting x0
- [ ] check eq 6
- [ ] check reference of eq 6 (wrongly ref as eq.1)
- [ ] use Differential Equations instead of DE
- [ ] remove \bm
- [ ] fix documentation test

rupof:
Hi! 

I implemented all the suggested changes:

- [x]  setting x0
- [x]  check eq 6
- [x]  check reference of eq 6 (wrongly ref as eq.1)
- [x]  use Differential Equations instead of DE
- [x]  remove \bm
- [x]  fix documentation test

Any other feedback, I am happy to help :)

David-Kreplin (commented):
There are a few rendering errors left:
- For the logistic differential equation, we take $ Q(x) = P(x) = -6 $ and $ n = 2$ in the Eq. 
-  This is, $:nbsphinx-math:mathcal{L}{:nbsphinx-math:`theta`} = :nbsphinx-math:`mathcal{L}`{\theta}(dy/dx, y, x) + \eta `:nbsphinx-math:left`(y(x_0) - u_0 \right)^2 $, where 

The problem is that you probably have no possibility to look at the final rendered HTML file; the notebook looks fine, on the other hand.

David-Kreplin (approved):
Loos great now, thank you!

PR #317, opened 2025-02-06
Add differential equation solver (High level QKODE) for Quantum Kernels

rupof:
Hi!

This is the implementation for the differential equation solver using QK (as done in https://arxiv.org/abs/2203.08884). It should accept linear and non-linear 1st order and 2nd order differential equations that depend on one variable. Furthermore, it should be able to support precomputed kernels  (i.e. classical kernels also work for it).

I also created an example (example_kode.ipynb) of how to use the interface, and modified+renamed the previous QNN ODE tutorial to include also QKs. 

Overall the implementation is somewhat similar to the QNN ODE solver.

Feel free to give me feedback and let me know what should be improved. 

Thank you very much for your help! 🚀
Roberto


src/squlearn/kernel/loss/ODE_loss.py
```diff
@@ -0,0 +1,225 @@
...
+import numpy as np
+import sympy as sp
+
+from .kernel_loss_base import KernelLossBase
+from ..lowlevel_kernel.kernel_matrix_base import KernelMatrixBase
+
+
+class ODELoss(KernelLossBase):
+    r"""
+    Ordinary Differential Equation (ODE) loss function for Quantum Kernels. It uses the same style as the QNN ODE loss function.
```
David-Kreplin, line 13: Shorten it to 100 charaters (ODE does not need to be explained in the oneliner)

examples/kernel/example_qkode.ipynb
David-Kreplin, line 1: This file looks broken to me :/
Maybe you can check it and fix it if possible

src/squlearn/kernel/loss/ODE_loss.py
```diff
@@ -0,0 +1,225 @@
...
+    def compute(
+        self,
+        parameter_values: np.ndarray,
+        data: np.ndarray,
+        labels: np.ndarray,
+        kernel_tensor: np.ndarray = None,  # [K, dKdx, dKdxdx] where dKdx is a np.ndarray of shape (n_samples, n_samples) and dKdxdx is a np.ndarray of shape (n_samples, n_samples)
+    ) -> float:
+        """
+        Calculates the squared loss of the loss function for the ODE as
+
```
David-Kreplin, line 171: Check the linewidth

src/squlearn/kernel/loss/ODE_loss.py
```diff
@@ -0,0 +1,225 @@
...
+            )
+        elif order_of_ODE == 2:
+            print(
+                "WARNING: 2nd order DEs differentiate the loss function by calculating the"
+                " second derivative. This can be computationally expensive and inneficient."
+            )
+        elif order_of_ODE > 2:
+            raise ValueError("Currently, only 1rst and 2nd order ODEs are supported")
+
+    def compute(
```
David-Kreplin, line 162: Is the gradient not needed?

David-Kreplin (changes requested):
Thank you very much, good progress so far!

I added QKODE and ODE_loss to the documentation. Unfortunately, this causes many errors when building the documentation.

The ode_solver.ipynb notebook needs to be restarted. I reduced the number of iterations to 3 to check if the documentation works. It looks fine, but the original cumbersome optimization needs to be restarted. Somehow, the original one has not been run either which caused a problem in the documentation build. 

Unfortunately, example_qkode seems broken. VS Code can only display the text version, but not the notebook.

The programming code itself looks fine :).

src/squlearn/kernel/qkode.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def fit(self, X, y, param_ini=None, K=None, dKdx=None, dKdxdx=None):
+        """ """
+        X, y = self._validate_data(
+            X, y, accept_sparse=("csr", "csc"), multi_output=True, y_numeric=True
+        )
+        self.X_train = X
+
+        # set up kernel matrix
+        if isinstance(self._quantum_kernel, str):
```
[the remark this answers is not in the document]
RobertoFlorez: Hmmm I dont know which ones because they take X and y as inputs. And X,y are only provided in the fit method

src/squlearn/kernel/qkode.py
```diff
@@ -0,0 +1,162 @@
...
+                self.k_train = K
+                self.dkdx_train = dKdx
+                if self._loss.order_of_ODE == 2:
+                    self.dkdxdx_train = dKdxdx
+            else:
+                raise ValueError("Unknown quantum kernel: {}".format(self._quantum_kernel))
+        elif isinstance(self._quantum_kernel, KernelMatrixBase):
+            # check if quantum kernel is trainable
+            if self._quantum_kernel.is_trainable:
+                print(
```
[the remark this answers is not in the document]
RobertoFlorez: In this implementation the parameters inside the quantum circuit are not optimized, only the classical parameters (i.e. alphas of the representer theorem, see other comment). I think implementing that second optimization (i.e. something like target alignment) could be possible but I have not done it.

src/squlearn/kernel/qkode.py
```diff
@@ -0,0 +1,162 @@
...
+                    "The Quantum Kernel is trainable but training the parameters of the kernel is not supported yet. Setting random parameters."
+                )
+                self._quantum_kernel.set_parameters(
```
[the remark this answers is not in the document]
RobertoFlorez: I agree :)

src/squlearn/kernel/qkode.py
```diff
@@ -0,0 +1,162 @@
...
+                self.dkdxdx_train = self._quantum_kernel.evaluate_derivatives(
+                    self.X_train, self.X_train, values="dKdxdx"
+                )
+
+        else:
+            raise ValueError(
+                "Unknown type of quantum kernel: {}".format(type(self._quantum_kernel))
+            )
+
+        if param_ini is None:
```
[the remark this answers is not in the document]
RobertoFlorez: I agree with the previous points that Parameter intialization should be done in the quantum_kernel, but here the parameters correspond to ansatz coefficients. Example:

Solving an ODE in the domain [0,0.5,1].

Then, the initial ansatz is: 

f(x) = alpha_0 + alpha_1 k(x, 0) + alpha_2 k(x, 0.5) + alpha_3 k(x, 1)

`param_ini`, here corrresponds to [alpha_0, alpha_1, alpha_2, alpha_3]

Perhaps the naming `param_ini` it is not ideal?

examples/kernel/example_qkode.ipynb
RobertoFlorez: Does it still look broken to you? In my computer it looks and run as a notebook fine

src/squlearn/kernel/loss/ODE_loss.py
RobertoFlorez: Hmmm I think this a good point. Right now, as I dont define a gradient method the optimizer uses FiniteDiffGradient. Originally, I had not implemented the gradient as the paper did not explicitly mention the need to implemented it(at least what I understood). However, probably it will improve the performance if I implemented, something like the implementation in `qnn.loss.ode_loss` but this will require an extra effort. 

If you think it is necessary for this PR, I could work on it. If not, I could make another PR or leave it as it is. 

What do you think?

RobertoFlorez:
Hi David, 

Thank you very much for the comments and sorry that it took me so long to answer (no rush to rereview :) ).

I implemented most of your suggestion and left a comment in the ones that were not trivial. 

`example_qkode` appears to be working properly on my computer. Does it still not display okay in your case?

Regarding `ode_solver.ipynb`, I reran it and pushed it but I still was unable to make it properly work with the documentation building. I think if we are unable to find the issue we could simple use the version that it is currently on the repo tutorials/qnn_ode_solver that does not include anything related to QKs. If you have any hints of where the documentation building problem could be, let me know and I can try to find the bug again. Sorry, that I was unable to find a solution for this.


Thanks again!

Best regards,
Roberto

src/squlearn/kernel/loss/ODE_loss.py
MoritzWillmann: we'll leave it as is for now

PR #319, opened 2025-02-06
Qulacs Integration

David-Kreplin:
First draft working for everything based on LowLevelQNNs (PQK, QNNs and QRC).

Currently this can provide speed-ups of 10x compared to the PennyLane implementation.

Todos:
- [x] Fidelity Kernel Implementation
- [x] Variance Implementation
- [x] Executor integration (caching)
- [x] Shots
- [x] Dependency Update


src/squlearn/util/qulacs/qulacs_circuit.py
```diff
@@ -0,0 +1,1029 @@
...
+    else:
+        obs_param_list = []
+
+    circ = circuit.get_circuit_func()(
+        *[
+            args[i]
+            for i, arg in enumerate(["param", "x"])
+            if arg in circuit._qualcs_gates_parameters
+        ]
+    )
```
MoritzWillmann, line 816: Hey @David-Kreplin, I made some changes here to allow for circuits without trainable parameters and features. I'm not entirely sure if this is still functioning as intended by you, so we should talk about it. I also think a solution using keyword args could be a lot cleaner and more robust. Same holds for all other evaluate functions later in the file

PR #321, opened 2025-03-24
Hot fix for Session creation

David-Kreplin:
Fix input arguments in session creation


src/squlearn/util/executor.py
```diff
@@ -2350,9 +2350,7 @@ def create_session(self):
 
         if self._service is not None:
```
MoritzWillmann, line 2351: do we still need this check (or even `self._service`) if we don't use it for session creation anymore?

David-Kreplin: Good point, probabaly self._service has become irrelevant

MoritzWillmann (approved):

PR #326, opened 2025-04-03
WIP: Qiskit 2 support

MoritzWillmann:
A couple of changes are necessary to support Qiskit 2.0. I started to adapt some things but there are still things missing, also some things wee need to wait on.

- [ ] `qiskit_algorithms` needs to be updated to support 2.0
- [ ] `BackendProperties` doesn't seem to exist anymore and needs to be replaced accordingly in `automatic_backend_selection`

### DIFF
--- a/src/squlearn/util/execution/automatic_backend_selection.py
+++ b/src/squlearn/util/execution/automatic_backend_selection.py
@@ -11,9 +11,6 @@ from packaging import version
 from qiskit import qasm3, transpile, QuantumCircuit
 from qiskit import __version__ as qiskit_version
 from qiskit.providers import Backend, BackendV2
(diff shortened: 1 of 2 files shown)


David-Kreplin:
Maybe a good chance to get rid of the qiskit_algorithms, which is only needed for fidelity quantum circuits. 
We have fast implementation for that for PennyLane and Qulacs, but not for Qiskit.

David-Kreplin:
Mapomatic causes problems with Qiskit 2.0 since it also checks for BackendV1 that is depricated in Qiskit 2.0.
I don't know, if they ever gonna update the Mapomatic Package, maybe we have to create a fix and hope someone will merge it and relase a new Mapomatic version.

MoritzWillmann:
@copilot can you please implement all comments from the review.

PR #331, opened 2025-04-29
Fix in Qiskit Encoding Circuit 

David-Kreplin:
Add copy statement, since this somehow causes problems with qiskit runtime.


MoritzWillmann (approved):

David-Kreplin:
Did not solve the problems I had and is therefore not necessary.

PR #339, opened 2025-07-15
model serialization

### LINKED ISSUE #275: Serialization of models
**Is your feature request related to a problem? Please describe.**
At the moment, there is no simple way for saving/loading of models. I have tried pickle, joblib, cloudpickle and none of them works.

**Describe the solution you'd like**
Enable the possibility of saving/loading of models with pickle or cloudpickle.

**Describe alternatives you've considered**
Alternatively, one could implement get_weights(), set_weights(weights: numpy.ndarray) for storing only the weights in the same way as in keras/tensorflow https://keras.io/guides/serialization_and_saving/

**Additional context**
Another solution is welcome!

DennisKleinhans:
## Changes Made
Introduce a custom serialization layer for QML models that cleanly strips out the `Executor` during pickling and re‑injects it on unpickling. This is achieved via:

### 1. `ExecutorPickler` / `ExecutorUnpickler`
- Uses Dill’s `persistent_id` / `persistent_load` hooks
- Replaces any `Executor` instance with a lightweight marker in the serialized stream
- Restores the actual `Executor` object at load time

### 2. `open_maybe` context manager
- Unifies file‑path and file‑like support
- Ensures correct opening/closing for disk files while allowing in‑memory streams

### 3. `SerializableModelMixin`
- Exposes `dump(target)` and `load(source, executor)` methods on any model
- Internally uses the custom pickler/unpickler and `open_maybe`
- Can be mixed into all existing model classes for seamless support

closes #275

### DIFF
--- a/src/squlearn/kernel/qgpc.py
+++ b/src/squlearn/kernel/qgpc.py
@@ -2,11 +2,13 @@
 
 from sklearn.gaussian_process import GaussianProcessClassifier
 
+from squlearn.util.serialization.serializable_model_mixin import SerializableModelMixin
+
 from .lowlevel_kernel.kernel_matrix_base import KernelMatrixBase
 from .lowlevel_kernel.kernel_util import kernel_wrapper
 
 
-class QGPC(GaussianProcessClassifier):
+class QGPC(GaussianProcessClassifier, SerializableModelMixin):
     """
     Quantum Gaussian process classification (QGPC), that extends the scikit-learn
     `sklearn.gaussian_process.GaussianProcessClassifier
--- a/src/squlearn/util/serialization/model_pickler.py
+++ b/src/squlearn/util/serialization/model_pickler.py
@@ -1,0 +1,35 @@
+import dill
+from squlearn.util.executor import Executor
+
+
+class ExecutorPickler(dill.Pickler):
+    """
+    Custom Pickler to handle the :ref:`Executor` object.
+    When an :ref:`Executor` is encountered, it will not be serialized directly.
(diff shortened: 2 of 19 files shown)


.github/workflows/doc_checks.yml
```diff
@@ -18,4 +18,4 @@ jobs:
           with:
             docs-folder: "docs/"
             pre-build-command: "pip3 install .[docs]"
-            build-command: "sphinx-build -b html . _build -W --keep-going"
+            build-command: "sphinx-build -b html . _build -W --keep-going -j 1"
```
MoritzWillmann, line 21: do we actually need this now?

src/squlearn/kernel/qgpc.py
```diff
@@ -167,14 +168,3 @@ def set_params(self, **params) -> None:
...
-    @property
-    def quantum_kernel(self) -> KernelMatrixBase:
-        """Returns quantum kernel"""
-        return self._quantum_kernel
-
-    @quantum_kernel.setter
-    def quantum_kernel(self, quantum_kernel: KernelMatrixBase):
-        """Sets quantum kernel"""
-        self._quantum_kernel = quantum_kernel
-        self.kernel = kernel_wrapper(quantum_kernel)
```
MoritzWillmann, line 180: Why was this removed?

src/squlearn/util/serialization/model_pickler.py
MoritzWillmann, line 1: Since the pickler has now functionality for `SparsePauliOp` we should maybe think about another name

MoritzWillmann (commented):
minor comments

PR #359, opened 2025-11-06
fix parameter initialization

MoritzWillmann:
Fix the parameter initialization in the `BaseQNN` to allow `partial_fit` to work properly. So far, parameter would have been overwritten with every call


MoritzWillmann:
@DennisKleinhans please add some check for this to the tests

ProfessorNova (approved):

PR #360, opened 2025-11-06
fix `num_encoding_slots` in `QiskitEncodingCircuit`

MoritzWillmann:
Fixes the error of no `num_encoding_slots` in `QiskitEncodingCircuit`


MoritzWillmann:
@DennisKleinhans please add some checks for this to the tests

PR #363, opened 2025-11-12
Add break condition choice to adam minimize

MoritzWillmann:
This PR allows to choose the break condition for the `Adam` optimizer's `minimize` function between
- `param_update` to break when the parameter update is below tol, which is the old behavior and the default
- `func_value` to break when the func value goes below tol

maxiter and maxiter_total will be kept in both cases.

### DIFF
--- a/src/squlearn/optimizers/adam.py
+++ b/src/squlearn/optimizers/adam.py
@@ -12,7 +12,7 @@ class Adam(OptimizerBase, SGDMixin):
     Possible options that can be set in the options dictionary are:
 
     * **tol** (float): Tolerance for the termination of the optimization (default: 1e-6)
-    * **condition** (str): Break when parameter update is below tol ('param_update') or when
+    * **break_condition** (str): Break when parameter update is below tol ('param_update') or when
       the function value ('func_value') goes below tol (default: 'param_update')
     * **lr** (float, list, np.ndarray, callable): Learning rate. If float, the learning rate is constant.
       If list or np.ndarray, the learning rate is taken from the list or array.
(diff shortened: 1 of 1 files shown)


src/squlearn/optimizers/adam.py
```diff
@@ -12,6 +12,8 @@ class Adam(OptimizerBase, SGDMixin):
     Possible options that can be set in the options dictionary are:
 
     * **tol** (float): Tolerance for the termination of the optimization (default: 1e-6)
+    * **condition** (str): Break when parameter update is below tol ('param_update') or when
```
ProfessorNova:
```suggestion
    * **break_condition** (str): Break when parameter update is below tol ('param_update') or when
```

src/squlearn/optimizers/adam.py
```diff
@@ -37,6 +39,9 @@ def __init__(self, options: dict = None, callback=default_callback) -> None:  #
             options = {}
 
         self.tol = options.get("tol", 1e-6)
+        self.condition = options.get("condition", "param_update")
+        if self.condition not in ["param_update", "func_value"]:
+            raise ValueError("Condition must be 'param_update' or 'func_value'.")
```
ProfessorNova:
```suggestion
        self.break_condition = options.get("break_condition", "param_update")
        if self.break_condition not in ["param_update", "func_value"]:
            raise ValueError("Break condition must be 'param_update' or 'func_value'.")
```

src/squlearn/optimizers/adam.py
```diff
@@ -141,8 +146,12 @@ def minimize(
                 self.callback(self.iteration, self.x, gradient, fval)
 
             # check termination
-            if np.linalg.norm(self.x - x_updated) < self.tol:
-                break
+            if self.condition == "func_value":  # func_value
```
ProfessorNova:
```suggestion
            if self.break_condition == "func_value":  # func_value
```

src/squlearn/optimizers/adam.py
ProfessorNova: Renaming `condition` to `break_condition` could make the meaning of the option more clear. Condition alone could be missunderstood as something like boundary condition.

PR #366, opened 2025-11-26
add func_update break condition to adam

MoritzWillmann:
adds and additional break condition for adam optimizer where the update of the loss function is checked

### DIFF
--- a/src/squlearn/optimizers/adam.py
+++ b/src/squlearn/optimizers/adam.py
@@ -13,8 +13,8 @@ class Adam(OptimizerBase, SGDMixin):
 
     * **tol** (float): Tolerance for the termination of the optimization (default: 1e-6)
     * **break_condition** (str): Break when parameter update is below tol ('param_update'), when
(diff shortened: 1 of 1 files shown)


src/squlearn/optimizers/adam.py
```diff
@@ -12,8 +12,9 @@ class Adam(OptimizerBase, SGDMixin):
     Possible options that can be set in the options dictionary are:
 
     * **tol** (float): Tolerance for the termination of the optimization (default: 1e-6)
-    * **break_condition** (str): Break when parameter update is below tol ('param_update') or when
-      the function value ('func_value') goes below tol (default: 'param_update')
+    * **break_condition** (str): Break when parameter update is below tol ('param_update'), when
+      the function value ('func_value') goes below tol or the function value does not change
+      ('func_update') (default: 'param_update').
```
ProfessorNova:
```suggestion
      the function value ('func_value') goes below tol or the function value change is below
      tol ('func_update') (default: 'param_update').
```

PR #390, opened 2026-06-15
Transpiled Circuit derivative

NicoRenaud:
Dear squlearn devs, 

thanks for the great library you are creating ! We are starting to use squlearn on IQM machines and we ran into some small issues. One of them is related to set of native gates statically defined [here](https://github.com/sQUlearn/squlearn/blob/0f6cb0982fdbe80dcd0ea6cb0acd68c59e362de4/src/squlearn/util/optree/optree_derivative.py#L352). Since IQM backends have a very limited set of native gates, this caused issues with qiskit runtime. 

To circumvent the issue, I've piped the backend info all the way to the optree class while trying to keep the original behavior as default.  It might be useful to also consider specifying the transpilation options a bit more dynamically and potentially use the TranspiledEncodingCircuit class instead of simply calling the transpile function.

If you think that's useful, I can put some more work into this as we need it for our project. However before doing so I would love to know what you think or if you see a simpler option to make squlearn compatible with IQM. 

Best wishes
Nico

### DIFF
--- a/src/squlearn/encoding_circuit/encoding_circuit_derivatives.py
+++ b/src/squlearn/encoding_circuit/encoding_circuit_derivatives.py
@@ -92,6 +92,7 @@ class EncodingCircuitDerivatives:
         encoding_circuit: EncodingCircuitBase,
         num_features: int,
         optree_caching: bool = True,
+        backend = None
     ):
         self.encoding_circuit = encoding_circuit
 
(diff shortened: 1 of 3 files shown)


David-Kreplin:
Dear Nico,

Thank you very much for pointing out the issues with the IQM machines. We have not yet had the chance to test this on the IQM hardware ourselves, so this kind of feedback is very helpful.

Regarding your commit: I think it may solve the problem for your particular case, but it could introduce another issue. The purpose of the transpilation step in the OpTree routine, which you modified, is to transform the circuit into a basis set for which we can implement the parameter-shift rule in a simple and efficient way. This is then handled automatically in the EncodingCircuitDerivatives class.

Your change might work well for the basis gates of the IQM machine, but it may not generalize to other hardware-specific basis gates.

I will think about a more general solution before merging the commit.

Thanks again, and best wishes,
David

NicoRenaud:
Dear David, 

thanks for your answer ! I'm simply trying the code with the fake IQM backends at the moment but it seems to be working fine. We are planning to run some experiments on real backend (specifically on the IQM machine installed at LUMI-Q). I'll let you know what your experience is on that backend. 

In the meantime let me know if I can do anything to help !

Best wishes
Nico

MoritzWillmann:
@NicoRenaud @David-Kreplin please make sure to base this pr onto develop not main! Thanks!

PR #392, opened 2026-07-06
WIP: Integrate `qc_executor` as sQUlearn's unified quantum backend

### LINKED ISSUE #391: Integration of `qc_executor`
# Replace sQUlearn's internal Executor with `qc_executor`

## Summary

sQUlearn currently maintains its own executor abstraction and three separate
low-level QNN implementations (`LowLevelQNNQiskit`, `LowLevelQNNPennyLane`,
`LowLevelQNNQulacs`), dispatched via a factory (`LowLevelQNN`) as well as Kernel implementations. We want to replace
the internal executor entirely with `qc_executor`, our shared framework for running
quantum circuits across Qiskit, PennyLane, and Qulacs, and consolidate the three
QNN implementations into a single one.

## Why this is tractable

Both sQUlearn and `qc_executor` use **Qiskit as their common internal representation**.
sQUlearn's encoding circuits and observables are built on top of Qiskit today, and
`qc_executor` natively understands Qiskit circuits (`QiskitCircuit`). This means we
don't need an explicit circuit-conversion layer — the circuit abstraction that used
to live in sQUlearn has effectively moved into `qc_executor`. sQUlearn's role becomes
a thin layer providing the circuit library, observable definitions, and the
QNN/Kernel user-facing API.

The `OpTree` structure — used today by `LowLevelQNNQiskit` to build derivative
expressions — already lives in `qc_executor`. We don't need to rebuild it; we just
import it from there.

`qc_executor` exposes: `expectation_value`, `expectation_value_derivatives`,
`sample`, `statevector`, and dispatches to the right backend (Qiskit / PennyLane /
Qulacs) internally.

## Target architecture

```
sQUlearn
├── Circuit Library / EncodingCircuitBase
│     get_circuit() → qc_executor.QuantumCircuitBase   (was: qiskit.QuantumCircuit)
├── ObservableBase
│     get_operator() → qc_executor.QuantumOperatorBase (was: qiskit operator)
├── EncodingCircuitDerivatives / ObservableDerivatives
│     unchanged, now import OpTree from qc_executor
└── LowLevelQNN (single class, no factory)
      calls qc_executor.expectation_value() / expectation_value_derivatives()

qc_executor (framework-agnostic, no sQUlearn awareness)
├── AbstractExecutor
│    ├── QiskitExecutor   (+ QiskitCircuit, QiskitOperator)
│    ├── PennyLaneExecutor
│    └── QulacsExecutor
└── OpTree, QuantumCircuitBase, QuantumOperatorBase
```

### Key design decision: connection point between `EncodingCircuitBase` and `qc_executor`

We considered having `EncodingCircuitBase` inherit from `qc_executor.QuantumCircuitBase`,
but rejected it: `QuantumCircuitBase` *is* an executable circuit instance (gates,
parameter assignment, backend representation), while `EncodingCircuitBase` *is a
factory* that produces a circuit via `get_circuit()` — it never holds circuit state
itself. Inheritance would imply the wrong semantics and burden `EncodingCircuitBase`
with a gate API it doesn't need.

Instead, the connection happens one level down, at the return type of `get_circuit()`
(and analogously `ObservableBase.get_operator()`):

```python
# Before
def get_circuit(self, features, parameters) -> QuantumCircuit:      # Qiskit

# After
def get_circuit(self, features, parameters) -> QuantumCircuitBase:  # qc_executor
```

This is a minimal-diff change for existing subclasses — they build the same Qiskit
circuit internally and just wrap it before returning.

### M1 — Connect `EncodingCircuitBase` to `qc_executor`
- [x] Change `get_circuit()` return type from `qiskit.QuantumCircuit` to
      `qc_executor.QuantumCircuitBase`
- [x] Update all existing `EncodingCircuitBase` subclasses (circuit library) accordingly
- [x] Unit tests: circuit output identical (structurally) before/after

### M2 — Connect `ObservableBase` to `qc_executor`
- [x] Change `get_operator()` return type to `qc_executor.QuantumOperatorBase`
- [x] Update all existing `ObservableBase` subclasses
- [x] Unit tests: operator output identical before/after

### M3 — Point `EncodingCircuitDerivatives` / `ObservableDerivatives` at `qc_executor`'s `OpTree`
- [x] Replace local `OpTree` import with `qc_executor.OpTree`
- [x] Confirm no behavioral differences in derivative tree construction

### M4 — Migrate `LowLevelQNNQiskit` to `qc_executor` (Strangler Fig)
- [x] Remove `_set_primitive()` / Estimator-Sampler selection, hold a `qc_executor`
      instance instead
- [x] Replace `OpTree.evaluate.evaluate_with_estimator/sampler(...)` calls with
      `qc_executor.expectation_value_derivatives(...)`
- [x] Keep the public interface of `LowLevelQNNQiskit` unchanged; factory and existing
      call sites remain untouched
- [x] Implement end-to-end tests covering the full pipeline
      (`EncodingCircuit -> LowLevelQNN -> qc_executor -> result`) as a safety net for
      this first full integration

### M5 — Migrate `LowLevelQNNPennyLane`
- [ ] Replace `qml.QNode` + `qml.jacobian` calls with `qc_executor` equivalents
- [ ] Regression tests against current PennyLane behavior

### M6 — Migrate `LowLevelQNNQulacs`
- [ ] Replace `qulacs_evaluate` / `qulacs_gradient` calls with `qc_executor` equivalents
- [ ] Remove the `_not_implemented` derivative-order allowlist — capability checks
      move to `qc_executor`
- [ ] Regression tests against current Qulacs behavior

### M7 — Consolidate into a single `LowLevelQNN`
- [ ] Extract shared logic (caching, output reshaping, parameter preprocessing —
      currently near-duplicated across all three classes) into one implementation
- [ ] Remove the `LowLevelQNN` factory; framework dispatch now lives entirely in
      `qc_executor`
- [ ] Delete `LowLevelQNNQiskit`, `LowLevelQNNPennyLane`, `LowLevelQNNQulacs`,
      and the old sQUlearn `Executor`

DennisKleinhans:
This pull request replaces sQUlearn's internal executor with `qc_executor`, establishing it as the single, uniform backend for circuit execution across different quantum frameworks.
As part of this integration:

- All coupling to sQUlearn's internal executor abstraction is removed from the codebase (encoding circuits, observables, and low-level QNN/kernel implementations now interface with qc_executor directly).
- The legacy executor implementation itself is deleted, along with any framework-specific dispatch logic it previously handled.
- Circuit and observable construction is aligned with `qc_executor`'s API and argument conventions.

closes #391

### DIFF
--- a/src/squlearn/__init__.py
+++ b/src/squlearn/__init__.py
@@ -3,7 +3,7 @@
 from .util import Executor
 from . import observables, encoding_circuit, kernel, optimizers, qnn, util
 
-__version__ = "0.11.1"
+__version__ = "0.11.2"
 
 __all__ = [
     "Executor",


DennisKleinhans:
The integration for the `EncodingCircuit` classes is now complete for the time being, but it surfaced a structural issue worth discussing.

Several `EncodingCircuit` implementations rely heavily on Qiskit's `QuantumCircuit` API directly — in particular, they use gates that aren't currently exposed by `qc_executor`'s own `QuantumCircuit` wrapper. In these cases, the abstraction breaks down, and we're forced to reach into the underlying Qiskit circuit directly (via `.qiskit_circuit`) to apply the gate.

This works for now, but it's a fragile shortcut: it silently assumes that `qc_executor`'s `QuantumCircuit` will always be backed by Qiskit internally. If that internal representation ever changes — for example, if `qc_executor` moves to a different native backend — the entire encoding circuit library would break, since we'd be bypassing the abstraction layer that's supposed to protect us from exactly that kind of change.

I think it's worth thinking through a proper solution here so that we never need to touch the underlying representation directly. Concretely, that likely means extending `qc_executor`'s `QuantumCircuit` API to cover the missing gates/operations, rather than relying on direct access as a workaround.

PR #393, opened 2026-07-07
WIP: Integrate `qc_executor` as sQUlearn's unified quantum backend

### LINKED ISSUE #391: Integration of `qc_executor`

DennisKleinhans:
This pull request replaces sQUlearn's internal executor with `qc_executor`, establishing it as the single, uniform backend for circuit execution across different quantum frameworks.
As part of this integration:

- All coupling to sQUlearn's internal executor abstraction is removed from the codebase (encoding circuits, observables, and low-level QNN/kernel implementations now interface with qc_executor directly).
- The legacy executor implementation itself is deleted, along with any framework-specific dispatch logic it previously handled.
- Circuit and observable construction is aligned with `qc_executor`'s API and argument conventions.

closes #391 

Note: Replaces #392, which was accidentally closed after a force-push to `develop` in my fork. No content was lost — this PR contains the same work, now on a proper feature branch.


DennisKleinhans:
The integration for the `EncodingCircuit` classes is now complete for the time being, but it surfaced a structural issue worth discussing.

Several `EncodingCircuit` implementations rely heavily on Qiskit's `QuantumCircuit` API directly — in particular, they use gates that aren't currently exposed by `qc_executor`'s own `QuantumCircuit` wrapper. In these cases, the abstraction breaks down, and we're forced to reach into the underlying Qiskit circuit directly (via `.qiskit_circuit`) to apply the gate.

This works for now, but it's a fragile shortcut: it silently assumes that `qc_executor`'s `QuantumCircuit` will always be backed by Qiskit internally. If that internal representation ever changes — for example, if `qc_executor` moves to a different native backend — the entire encoding circuit library would break, since we'd be bypassing the abstraction layer that's supposed to protect us from exactly that kind of change.

I think it's worth thinking through a proper solution here so that we never need to touch the underlying representation directly. Concretely, that likely means extending `qc_executor`'s `QuantumCircuit` API to cover the missing gates/operations, rather than relying on direct access as a workaround.

jordan0809:
> The integration for the `EncodingCircuit` classes is now complete for the time being, but it surfaced a structural issue worth discussing.
> 
> Several `EncodingCircuit` implementations rely heavily on Qiskit's `QuantumCircuit` API directly — in particular, they use gates that aren't currently exposed by `qc_executor`'s own `QuantumCircuit` wrapper. In these cases, the abstraction breaks down, and we're forced to reach into the underlying Qiskit circuit directly (via `.qiskit_circuit`) to apply the gate.
> 
> This works for now, but it's a fragile shortcut: it silently assumes that `qc_executor`'s `QuantumCircuit` will always be backed by Qiskit internally. If that internal representation ever changes — for example, if `qc_executor` moves to a different native backend — the entire encoding circuit library would break, since we'd be bypassing the abstraction layer that's supposed to protect us from exactly that kind of change.
> 
> I think it's worth thinking through a proper solution here so that we never need to touch the underlying representation directly. Concretely, that likely means extending `qc_executor`'s `QuantumCircuit` API to cover the missing gates/operations, rather than relying on direct access as a workaround.

Is it possible to transpile the gates in `EncodingCircuit` so that all the gates are compatible with `qc_executor`'s `QuantumCircuit`? I think this might be more viable if we want to integrate `qc_executor` into other packages that might have  some other special gate sets or operations.

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -151,7 +150,7 @@ class _Id_operation(_operation):
 
     def get_circuit(self, var_param_assignment=None):
         QC = QuantumCircuit(self.num_qubits)
-        QC.id(range(self.num_qubits))
+        QC.qiskit_circuit.id(range(self.num_qubits))
```
jordan0809: Since it's the identity gate, we don't need to do anything except for returning `QC` right?

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -169,7 +168,7 @@ class _S_conjugate_operation(_operation):
 
     def get_circuit(self, var_param_assignment=None):
         QC = QuantumCircuit(self.num_qubits)
-        QC.sdg(range(self.num_qubits))
+        QC.qiskit_circuit.sdg(range(self.num_qubits))
```
jordan0809: I think there is `sdag` gate in `qc_executor`'s `QuantumCircuit`?

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -187,7 +186,7 @@ class _T_conjugate_operation(_operation):
 
     def get_circuit(self, var_param_assignment=None):
         QC = QuantumCircuit(self.num_qubits)
-        QC.tdg(range(self.num_qubits))
+        QC.qiskit_circuit.tdg(range(self.num_qubits))
```
jordan0809: Also `tdag`

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -972,14 +971,15 @@ def get_circuit(self, *args):
                 for i in range(operation_layer.num_layers):
                     for op in operation_layer.layer.operation_list:
                         if op.variablegroup_tuple == None:
-                            QC = QC.compose(op.get_circuit())
+                            QC = QC.compose(op.get_circuit(), None)
+                            QC.qiskit_circuit.num_qubits
```
jordan0809: What does this line do? I think we could also just use `QC.num_qubits`?

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -972,14 +971,15 @@ def get_circuit(self, *args):
                 for i in range(operation_layer.num_layers):
                     for op in operation_layer.layer.operation_list:
                         if op.variablegroup_tuple == None:
-                            QC = QC.compose(op.get_circuit())
+                            QC = QC.compose(op.get_circuit(), None)
```
jordan0809: The second argument should be `qubits: List[int]` or? (same goes for all other lines)

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -2186,16 +2186,16 @@ def generate_initial_parameters(self, num_features, seed=None):
...
 
         Args:
-            features Union[ParameterVector,np.ndarray]: Input vector of the features
+            features Union[Parameters,np.ndarray]: Input vector of the features
                 from which the gate inputs are obtained
-            param_vec Union[ParameterVector,np.ndarray]: Input vector of the parameters
+            param_vec Union[Parameters,np.ndarray]: Input vector of the parameters
                 from which the gate inputs are obtained
 
         Return:
```
jordan0809, line 2230: The docstring should be : `Returns the circuit in qc_executor QuantumCircuit format`

src/squlearn/encoding_circuit/pruned_encoding_circuit.py
```diff
@@ -1,11 +1,9 @@
 import numpy as np
-from qiskit.circuit import ParameterVector
-from qiskit.circuit.parametervector import ParameterVectorElement
+from qc_executor import Parameter, Parameters, QuantumCircuit
```
jordan0809: Correct me if I'm wrong, but I think we cannot import `Parameter` from `qc_executor` because its `__init__.py` file only has `from .parameters import Parameters`

src/squlearn/encoding_circuit/circuit_library/qcnn_encoding_circuit.py
```diff
@@ -676,16 +676,16 @@ def QCNNObservable(self, obs: Union[ObservableBase, str] = "Z") -> ObservableBas
...
-            param = ParameterVector("p", quantum_circuit.num_parameters)
+        if not isinstance(quantum_circuit, QiskitQuantumCircuit):
+            param = Parameters("p", quantum_circuit.num_parameters)
             if quantum_circuit.num_features > 0:
                 raise ValueError(
                     "No features are allowed in the QCNN ansatz. "
                     "Please provide a circuit without features instead."
                 )
-            quantum_circuit = quantum_circuit.get_circuit([], param)
+            quantum_circuit = quantum_circuit.get_circuit([], param).qisklit_circuit
```
jordan0809: why does it have to return `QiskitQuantumCircuit` here instead of  `qc_executor`'s `QuantumCircuit`

tests/encoding_circuit/test_layered_encoding_circuit.py
```diff
@@ -51,7 +50,7 @@ def test_layered_encoding_circuit_gates(self):
 
         # Test the Sdg gate
         lfm.S_conjugate()
-        expected_circuit.sdg(range(4))
+        expected_circuit.qiskit_circuit.sdg(range(4))
```
jordan0809, line 53: same issue as the src file. maybe just `expected_circuit.sdag(range(4))`?

src/squlearn/encoding_circuit/pruned_encoding_circuit.py
DennisKleinhans: Yes, you are right. However, I made a local change to the qc_executor so that I could use it. Clearly, this is something that should be changed in the qc_executor. There may be other things that come up that we need to change while the integration is ongoing.

src/squlearn/encoding_circuit/layered_encoding_circuit.py
DennisKleinhans: In theory, that's right. However, if someone wants to add an ID gate, we should do so as to not violate the expectation of having an added gate.

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -186,67 +181,62 @@ class _T_conjugate_operation(_operation):
...
+    def apply_param_vectors(self, QC: QuantumCircuit, var_param_assignment: dict):
         """
         Applies the param vectors creating a quantum circuit with the given gate.
         Contains an algorithm for Rx, Ry and Rz gates, which distinguish between the following cases:
         First: The user sets no map so a default map is given and the number of variable groups used exceeds 2, than raise an error because the default map needs exactly two arguments.
-        Second: The user sets no map so a default map is given and the number of variable groups used is 1, than apply the r* (stands for rx, ry or rz) gate with the given variable but without the map.
+        Second: The user sets no map so a default map is given and the number of variable groups used is 1, than apply the gate_method (stands for rx, ry or rz) gate with the given variable but without the map.
         Third: The user sets a map with some variable groups or there are exactly 2 variable groups given.
         Args:
             QC: the quantum circuit by qiskit
```
jordan0809: the quantum circuit by `qc-executor`?

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -376,16 +386,18 @@ def apply_param_vectors(self, QC, gate, var_param_assignment: dict):
         3rd: Variable group is given
         4th: Dimension (size) is finite (or infinite)
         """
-        if self.variablegroup_tuple == None:
+        gate_method = getattr(QC, self.gate_name)
```
jordan0809, line 399: What if the specified gate is not part of the `QuantumCircuit` methods? Should we raise an error here?
(Just saw that this function is only called for specific entangling operations defined in `QuantumCircuit`, but maybe raising an error will be helpful for future debugging if we add some new encoding operations here?)

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -687,16 +715,12 @@ def apply_param_vectors(self, QC, cu_gate, var_param_assignment):
                                 ]
                             )
                         variablegroup.increase_index(1)
-                    QC.qiskit_circuit.append(
-                        cu_gate(*buffer_param_vectors_list),
-                        [first_qubit, second_qubit],
-                        [],
-                    )
+                    QC.cu(first_qubit, second_qubit, *buffer_param_vectors_list)
```
jordan0809, line 728: I think `cu_gate` is not yet defined in `qc-executor` right?

src/squlearn/encoding_circuit/layered_encoding_circuit.py
```diff
@@ -970,16 +986,20 @@ def get_circuit(self, *args):
                 operation_layer = operation
                 for i in range(operation_layer.num_layers):
                     for op in operation_layer.layer.operation_list:
-                        if op.variablegroup_tuple == None:
-                            QC = QC.compose(op.get_circuit(), None)
-                            QC.qiskit_circuit.num_qubits
+                        if op.variablegroup_tuple is None:
+                            QC = QC.compose(op.get_circuit(), list(range(QC.num_qubits)))
+                            QC.num_qubits
```
jordan0809: Did you mean to log/print `QC.num_qubits` here, or can this line be removed?

src/squlearn/encoding_circuit/circuit_library/qcnn_encoding_circuit.py
```diff
@@ -483,6 +484,7 @@ def get_circuit(
                     qc_out.name = qc_name
                     if not gate[4]:
                         i_param += n_params
+                    # TODO: `circuit_to_gate` is qiskit specific. qc-executor should have a similar function to convert a circuit to a gate.
```
jordan0809: Maybe we don't need the `circuit_to_gate function`. The `compose` method in `qc_executor` already concatenates two `QuantumCircuit` objects, so we don't have to convert a circuit into a gate before composing it.

DennisKleinhans:
Quick note on `LowLevelQNNUnified`: it computes the first-order quantities (`f`, `dfdx`, `dfdp`, `dfdop`) directly via `qc_executor`, for all three frameworks (Qiskit, PennyLane, Qulacs). Everything beyond that (`dfdxdx`, `laplace`, `var`/`OO`, etc.) still delegates to the old framework-specific class (`LowLevelQNNQiskit`/`PennyLane`/`Qulacs`), lazily instantiated as a fallback on first use. This is an incremental step, not the end state, it's a bridge until `qc_executor` is fully integrated and the remaining parts of sQUlearn (higher-order derivatives, the rest of the `Executor` class, etc.) have been migrated too.

src/squlearn/qnn/lowlevel_qnn/lowlevel_qnn_pennylane.py
```diff
@@ -408,7 +420,9 @@ def _initialize_pennylane_circuit(self, num_features: int) -> None:
 
         self._x = ParameterVector("x", num_features)
         self._param = ParameterVector("param", self._pqc.num_parameters)
-        self._qiskit_circuit = decompose_to_std(self._pqc.get_circuit(self._x, self._param))
+        self._qiskit_circuit = decompose_to_std(
+            self._pqc.get_circuit(self._x, self._param).qiskit_circuit
```
jordan0809: Maybe we should pass `QuantumCircuit` directly into `decompose_to_std` instead of the qiskit circuit? This implies refactoring the  `src/squlearn/util/decompose_to_std.py` script. The resulting output will also be a `QuantumOperator`, so the variable should be renamed accordingly. Besides, `PennylaneCircuit` in `src/squlearn/util/pennylane/pennylane_circuit.py` should take `QuantumCircuit` and `QuantumOperator` as arguments, which will be in the future commits I suppose?

src/squlearn/qnn/lowlevel_qnn/lowlevel_qnn_qulacs.py
```diff
@@ -139,7 +151,9 @@ def _initialize_qulacs_circuit(self, num_features: int):
         # Parameter vectors for the PQC and the observable
         self._x = ParameterVector("x", num_features)
         self._param = ParameterVector("param", self._pqc.num_parameters)
-        self._qiskit_circuit = decompose_to_std(self._pqc.get_circuit(self._x, self._param))
+        self._qiskit_circuit = decompose_to_std(
+            self._pqc.get_circuit(self._x, self._param).qiskit_circuit
```
jordan0809: Also here if we decide to replace `qiskit_circuit`  with just `QuantumCircuit`

src/squlearn/qnn/lowlevel_qnn/lowlevel_qnn_pennylane.py
```diff
@@ -378,7 +380,11 @@ def _preprocess_observable(self) -> None:
             self._num_parameters_observable = self._observable.num_parameters
             self._param_obs = ParameterVector("param_obs", self._num_parameters_observable)
             self._qiskit_observable = self._observable.get_operator(self._param_obs)
```
jordan0809, line 381: I think this is actually the `qc_executor`'s `QuantumOperator` instead of qiskit operator now. Should we rename?

src/squlearn/qnn/lowlevel_qnn/lowlevel_qnn_pennylane.py
```diff
@@ -408,7 +420,9 @@ def _initialize_pennylane_circuit(self, num_features: int) -> None:
 
         self._x = ParameterVector("x", num_features)
         self._param = ParameterVector("param", self._pqc.num_parameters)
-        self._qiskit_circuit = decompose_to_std(self._pqc.get_circuit(self._x, self._param))
+        self._qiskit_circuit = decompose_to_std(
+            self._pqc.get_circuit(self._x, self._param).qiskit_circuit
+        )
 
         # PennyLane Circuit function of the QNN
         self._pennylane_circuit = PennyLaneCircuit(self._qiskit_circuit, self._qiskit_observable)
```
jordan0809, line 427: As mentioned above, `PennylaneCircuit` should be refactored. Currently, `self._qiskit_circuit` is still a qiskit circuit while `self._qiskit_observable` is actually a `QuantumOperator`.

src/squlearn/qnn/lowlevel_qnn/lowlevel_qnn_qulacs.py
```diff
@@ -139,7 +151,9 @@ def _initialize_qulacs_circuit(self, num_features: int):
...
+        )
 
         # Qulacs Circuit data structure of the QNN
         self._qulacs_circuit = QulacsCircuit(self._qiskit_circuit, self._qiskit_observable)
```
jordan0809, line 159: Same issue as the pennylane commit

jordan0809 (commented):
accidental post. Please ignore

DennisKleinhans:
All execution now goes through qc_executor. sQUlearn's own `Executor` is left with orchestration and the pieces that are genuinely its own: caching, retries, session bookkeeping, automatic backend selection, and QPU parallelization. No framework-specific execution or circuit-conversion logic remains on the sQUlearn side.

src/squlearn/qnn/lowlevel_qnn/lowlevel_qnn.py
```diff
@@ -1,51 +1,984 @@
...
+                native_observables.append(obs.get_operator(self._p_op[ioff:]))
+                self._observable_p_op_slices.append((ioff, obs.num_parameters))
+                ioff += obs.num_parameters
+            self._native_observable = native_observables
+            # <O^2> for the "var" family (see _VAR_FAMILY). Composing an observable with
+            # itself reuses the same p_op Parameter objects on both sides, so simplify()
+            # cannot introduce new parameter symbols beyond the ones already in
+            # _observable_p_op_slices
+            self._native_observable_squared = [
+                QuantumOperator(_native_operator=obs.qiskit_operator)
```
jordan0809, line 243: I think `_native_operator` is not a keyword argument supported by `QuantumOperatorBase` and it's also private. Maybe use `from_quantum_operator`?

tests/util/test_executor.py
```diff
@@ -300,19 +293,14 @@ def test_shots(self, executor_str, request):
...
-            "ExecutorPennyLane": np.array([0.0, 0.0, 0.0, 1.0]),
-            "ExecutorPennyLaneShots": np.array([0.0, 0.0, 0.0, 1.0]),
-            "ExecutorPennyLaneDevice": np.array([0.0, 0.0, 0.0, 1.0]),
-        }
+        """Executor.probabilities() through the native qc_executor path - the
+        successor to the removed Executor.pennylane_execute()."""
 
         executor = request.getfixturevalue(executor_str)
-        circuit = PennyLaneCircuit(simple_circuit, "probs")
+        circuit = QcExecutorQuantumCircuit.from_qiskit(simple_circuit)
```
jordan0809, line 300: `.from_qiskit` is not supported by `QuantumCircuit`. Same goes for all other tests.

tests/encoding_circuit/circuit_library/test_qiskit_encoding_circuit.py
```diff
@@ -1,24 +1,25 @@
...
         qc_out = circuit.get_circuit([], [])
-        assert qc_out is circuit._qiskit_circuit
+        assert qc_out.qiskit_circuit is circuit._qiskit_circuit
 
     def test_mode_x_assigns_features_and_raises_on_wrong_length(self):
-        params = ParameterVector("x", 2)
-        qc = QuantumCircuit(2)
+        params = Parameters("x", 2)
+        qc = QiskitQuantumCircuit(2)
         qc.rz(params[0], 0)
```
jordan0809, line 23: Just a reminder that in `framework-independent-ir`, `params[0]` is sympy-based and is therefore incompatible with `QiskitQuantumCircuit`.

tests/encoding_circuit/circuit_library/test_chebyshev_pqc.py
jordan0809, line 24: A reminder that in `framework-independent-ir`, `Parameter` is sympy-based and would not be accepted in a numpy function like `np.arccos(x)`

